PR #50, opened 2026-08-19
HTTP transport handling improvements

niello:
Solved some problems with the current upstream implementation:

- Better handling of transport parameter:
  - setting via FASTMCP_TRANSPORT environment variable
  - consistent `http` -> `streamable-http` aliasing
  - `stdio` by default without explicit docker CMD
- supported `TransportSecuritySettings` present in newer FastMCP versions according to the guide at https://github.com/modelcontextprotocol/python-sdk/issues/1798. Solves "OPTIONS /mcp HTTP/1.1" 421 Misdirected Request.
- added CORS middleware. Solves "OPTIONS /mcp HTTP/1.1" 405 Method Not Allowed.

I didn't find a contribution guide, so I did my best to explain problems and changes above. Feel free to ask if you need more information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -760,6 +760,41 @@ docker run --rm -p 8000:8000 \
 - Remote HTTP is typically **unauthenticated** and **unencrypted** by default; don’t expose this port publicly. Use VPN/firewall rules or a reverse proxy with TLS + auth.
 - Don’t bake API keys into the image; pass them via env vars at runtime.
 
+### Host and origin allowlists (`421` and `403` errors)
+
+The HTTP transports enforce DNS rebinding protection. Out of the box only loopback is
+accepted — `127.0.0.1`, `localhost` and `[::1]` — which covers `docker run -p 8000:8000`
+reached as `http://localhost:8000/mcp`, but not much else. Reach the server under any
+other name (a Compose service name, a LAN address, a reverse-proxy hostname) and you get:
+
+- `421 Invalid Host header` — the `Host` header isn’t in the allowlist.
+- `403 Invalid Origin header` — a browser-based client sent an `Origin` that isn’t.
+
+Widen the allowlist with two comma-separated variables:
+
+| Variable | Matches against | Example |
+|---|---|---|
+| `FASTMCP_ALLOWED_HOSTS` | the `Host` header | `kindly-web-search-mcp:*,localhost:*,127.0.0.1:*` |
+| `FASTMCP_ALLOWED_ORIGINS` | the `Origin` header (browser clients) | `http://localhost:*,https://app.example` |
+
+Syntax notes:
+
+- An entry is either an **exact** value or a `name:*` **port wildcard**. A bare `*` matches
+  nothing — `FASTMCP_ALLOWED_HOSTS=*` rejects every request.
+- `kindly-web-search-mcp:*` matches `kindly-web-search-mcp:8000` but **not** a portless
+  `Host: kindly-web-search-mcp`. List both if your client may omit the port.
+- Setting a variable **replaces** the loopback defaults for that header rather than adding
+  to them. Keep `localhost:*` and `127.0.0.1:*` in the list if you still want local access.
+- The two default independently. Setting only `FASTMCP_ALLOWED_HOSTS` keeps the loopback
+  origins, so browser clients on `http://localhost:<port>` keep working.
+- CORS is derived from `FASTMCP_ALLOWED_ORIGINS`, so an origin that passes preflight is
+  always one the server will also accept on the real request.
+
+Neither variable is a way to turn the protection off, and that is deliberate: this server
+is unauthenticated, and `get_content` will fetch any URL it is given from wherever it runs.
+An open allowlist lets any web page you happen to visit drive it against your own network
+and read the response. List the hosts you actually use.
+
 ### Docker-compose
 
 Add a subsection to `services:` in your `docker-compose.yml`. Setup variables as described above.
@@ -778,9 +813,14 @@ Add a subsection to `services:` in your `docker-compose.yml`. Setup variables as
       - FASTMCP_HOST=0.0.0.0
       - FASTMCP_PORT=8000
       - FASTMCP_TRANSPORT=http
+      # Other containers reach this one by its service name, which is not loopback.
+      # Without this you get `421 Invalid Host header`. See the section above.
+      - FASTMCP_ALLOWED_HOSTS=kindly-web-search-mcp:*,localhost:*,127.0.0.1:*
 ```
 
 `FASTMCP_TRANSPORT` can be `streamable-http` (or simply `http`) for Streamable HTTP, `sse` for SSE.
+It is read by every entry point, including `kindly-web-search-mcp-server start-mcp-server`.
+An unrecognised value logs a warning and falls back to stdio.
 
 Run with:
 
@@ -789,6 +829,11 @@ docker compose up -d
 ```
 
 Container will be built at the first run. To rebuild it, append `--build` to the command above.
+
+> **Warning:** `ports: - "8000:8000"` publishes the server on **all** host interfaces. As with
+> `docker run` above, remote HTTP here is **unauthenticated** and **unencrypted** — don’t expose
+> this port publicly. Bind it to loopback (`"127.0.0.1:8000:8000"`), keep it on an internal
+> Compose network, or put a reverse proxy with TLS + auth in front.
 
 ## Troubleshooting
 
--- a/README.md
+++ b/README.md
@@ -760,6 +760,36 @@ docker run --rm -p 8000:8000 \
 - Remote HTTP is typically **unauthenticated** and **unencrypted** by default; don’t expose this port publicly. Use VPN/firewall rules or a reverse proxy with TLS + auth.
 - Don’t bake API keys into the image; pass them via env vars at runtime.
 
+### Docker-compose
+
+Add a subsection to `services:` in your `docker-compose.yml`. Setup variables as described above.
+
+```yaml
+  kindly-web-search-mcp:
+    build:
+      context: https://github.com/Shelpuk-AI-Technology-Consulting/kindly-web-search-mcp-server.git#main
+    container_name: kindly-web-search-mcp
+    restart: unless-stopped
+    ports:
+      - "8000:8000"
+    environment:
+      - SEARXNG_BASE_URL=...
+      - GITHUB_TOKEN=${KINDLY_GITHUB_TOKEN:-}
+      - FASTMCP_HOST=0.0.0.0
+      - FASTMCP_PORT=8000
+      - FASTMCP_TRANSPORT=http
+```
+
+`FASTMCP_TRANSPORT` can be `streamable-http` (or simply `http`) for Streamable HTTP, `sse` for SSE.
+
+Run with:
+
+```bash
+docker compose up -d
+```
+
+Container will be built at the first run. To rebuild it, append `--build` to the command above.
+
 ## Troubleshooting
 
 - “No Chromium-based browser executable found”: install Chrome/Chromium/Edge and set `KINDLY_BROWSER_EXECUTABLE_PATH` if needed.
--- a/src/kindly_web_search_mcp_server/server.py
+++ b/src/kindly_web_search_mcp_server/server.py
@@ -101,9 +101,12 @@ def _build_arg_parser() -> argparse.ArgumentParser:
 
 
 def _resolve_transport(raw: str | None) -> Transport:
-    if raw in ("stdio", "sse", "streamable-http"):
-        return raw
-    return os.environ.get("FASTMCP_TRANSPORT", "stdio")
+    resolved_transport = raw or os.environ.get("FASTMCP_TRANSPORT", "stdio")
+    if resolved_transport == "http":
+        return "streamable-http"
+    if resolved_transport in ("stdio", "sse", "streamable-http"):
+        return resolved_transport
+    return "stdio"
 
 
 def _resolve_host_port(host: str | None, port: int | None) -> tuple[str, int]:
--- a/src/kindly_web_search_mcp_server/server.py
+++ b/src/kindly_web_search_mcp_server/server.py
@@ -4,6 +4,7 @@ import argparse
 import asyncio
 import logging
 import os
+import re
 import sys
 import time
 from typing import Literal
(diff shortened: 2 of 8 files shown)


shelpuk (commented):
Thanks for this — the problems you describe are real, and I reproduced both of them before reviewing. I checked out the branch, installed it, and ran it side by side against `main`:

| Check | `main` | this PR |
|---|---|---|
| `POST /mcp` with `Host: kindly-web-search-mcp:8000` | `421 Invalid Host header` | `200` |
| `OPTIONS /mcp` from a browser origin | `403` | `200`, `Access-Control-Allow-Origin: *` |

So the diagnosis is accurate and the general shape (`transport_security` + CORS middleware) is the right fix. There are three things I'd like resolved before merging, plus some smaller notes. Details are in the inline comments; summarising here:

**Blocking**

1. **`--sse` is silently broken.** The `hasattr` check always picks `streamable_http_app()`, so `--sse` / `FASTMCP_TRANSPORT=sse` now serves `/mcp` and `/sse` returns 404.
2. **Security regression in the default configuration.** With no `FASTMCP_ALLOWED_*` set — the default for every existing user — DNS rebinding protection is off *and* CORS is `*`. I verified a page on `https://evil.example` can complete an MCP `initialize` and read the response; `main` rejects it with 403.
3. **Setting only `FASTMCP_ALLOWED_HOSTS` produces a broken half-state**: preflight returns 200 with `*`, the real request returns 403.

**Non-blocking**

- `--mount-path` is now ignored on the HTTP path.
- `uvicorn` and `starlette` are imported directly but aren't declared in `pyproject.toml` — they only arrive transitively via `mcp`. `tests/test_dependency_constraints.py` exists to guard exactly this class of problem; please add them to `[project].dependencies`.
- No tests. I ran the suite on both branches and it's identical (11 failed, 140 passed — those 11 are pre-existing Windows/nodriver failures, unrelated to this PR), so nothing new breaks, but nothing covers the new behaviour either. I'd like tests for `_resolve_transport` (env var, `http` alias, CLI-over-env precedence, invalid-value fallback) and one asserting `--sse` still serves `/sse`.
- `kindly-web-search-mcp-server start-mcp-server` injects `--stdio` when no transport flag is present (`src/kindly_web_search_mcp_server/cli.py:51`), so `FASTMCP_TRANSPORT` is ignored on that entry point. Worth either honouring it there too or noting the exception in the README.

The `Dockerfile` change is correct — dropping `CMD` is what lets the env var take effect, and the default stays stdio. No concerns there.

Re: the contribution guide — there isn't one yet, that's on us, not you. The house rules that matter here are: every new behaviour gets a test, and public functions get Google-style docstrings. Happy to pair on the SSE and default-security bits if you'd rather not carry them alone.

src/kindly_web_search_mcp_server/server.py
```diff
@@ -176,11 +190,28 @@ def main(argv: list[str] | None = None) -> None:
...
-    try:
-        mcp.run(transport=transport, mount_path=args.mount_path)
-    except TypeError:
-        # Backward-compat: older MCP SDKs may not accept `mount_path`.
-        mcp.run(transport=transport)
+        asgi_app = (
+            mcp.streamable_http_app()
+            if hasattr(mcp, "streamable_http_app")
+            else mcp.sse_app()
+        )
```
shelpuk, line 300: **Blocking: this breaks the `sse` transport.**

`streamable_http_app` exists on every SDK version this project supports (`mcp>=1.25,<2` — I checked both 1.25.0 and 1.29.1), so the `sse_app()` branch is unreachable. The result is that `--sse` and `FASTMCP_TRANSPORT=sse` silently serve Streamable HTTP instead.

Verified on both branches:

```
main --sse:  GET /sse -> stream held open (works);  /mcp -> 404
this --sse:  GET /sse -> 404;                       /mcp -> 200
```

The route sets differ, so this isn't cosmetic — existing SSE clients will get a 404 after upgrading:

```python
streamable_http_app routes: ['/mcp']
sse_app routes:            ['/sse', '/messages']
```

Please branch on the resolved transport rather than on `hasattr`:

```python
asgi_app = mcp.sse_app(args.mount_path) if transport == "sse" else mcp.streamable_http_app()
```

That also restores `--mount-path`, which is otherwise dropped on this path (see the note on the `else` branch below).

src/kindly_web_search_mcp_server/server.py
```diff
@@ -26,12 +29,20 @@
...
     "kindly-web-search",
     instructions=(
         "Web search via Serper (default), Tavily, or a self-hosted SearXNG instance with best-effort "
         "scraping/extraction of result pages into Markdown for LLM consumption."
     ),
+    transport_security=TransportSecuritySettings(
+        enable_dns_rebinding_protection=bool(allowed_hosts or allowed_origins),
+        allowed_hosts=allowed_hosts,
+        allowed_origins=allowed_origins,
+    ),
```
shelpuk: **Blocking: this disables DNS rebinding protection for everyone by default, not just for compose users.**

I think there's a false premise here. The SDK's default is *not* "no settings" — `FastMCP` already ships a populated `TransportSecuritySettings`:

```
enable_dns_rebinding_protection=True
allowed_hosts=['127.0.0.1:*', 'localhost:*', '[::1]:*']
allowed_origins=['http://127.0.0.1:*', 'http://localhost:*', 'http://[::1]:*']
```

(Same on 1.25.0 and 1.29.1.) That default is exactly what produces your 421 when reaching the container by its compose service name — correctly diagnosed. But `bool(allowed_hosts or allowed_origins)` replaces a localhost allowlist with *no restriction at all* whenever the env vars are unset, which is the default for every current user of this server.

Combined with `allow_origins=["*"]` below, I confirmed that a page served from `https://evil.example` can drive the local server:

```
main:      POST /mcp, Origin: https://evil.example -> 403 Invalid Origin header
this PR:   POST /mcp, Origin: https://evil.example -> 200, full initialize result readable
```

That matters more than usual here because the server is unauthenticated and exposes `get_content` (fetches an arbitrary URL server-side — SSRF into the user's LAN, with the response body readable by the attacking page) and `web_search` (spends the user's paid provider keys).

Suggested shape: leave the SDK default in place unless the operator opts in, and make relaxing it explicit.

```python
transport_security = (
    TransportSecuritySettings(
        enable_dns_rebinding_protection=True,
        allowed_hosts=allowed_hosts,
        allowed_origins=allowed_origins,
    )
    if allowed_hosts or allowed_origins
    else None  # keep the SDK's localhost-only default
)
```

and pass `transport_security` only when it isn't `None`. If you also want a genuine escape hatch for "any host", make it a separate, deliberately-named opt-out rather than the behaviour you fall into by not configuring anything.

One documentation detail worth capturing wherever this lands: the SDK's wildcard matcher only handles a `host:*` port suffix (`_validate_host` checks exact match, then `allowed.endswith(":*")`). A bare `*` does **not** match anything. So the value that actually works for the compose example is:

```
FASTMCP_ALLOWED_HOSTS=kindly-web-search-mcp:*,localhost:*
```

src/kindly_web_search_mcp_server/server.py
```diff
@@ -176,11 +190,28 @@ def main(argv: list[str] | None = None) -> None:
...
+        asgi_app = (
+            mcp.streamable_http_app()
+            if hasattr(mcp, "streamable_http_app")
+            else mcp.sse_app()
+        )
+
+        # NB: allow_credentials=True will require explicit methods & headers
+        app = CORSMiddleware(
+            asgi_app,
+            allow_origins=allowed_origins if allowed_origins else ["*"],
```
shelpuk: **Blocking (related to the `transport_security` comment): `["*"]` as the fallback, plus a confusing half-configured state.**

Two issues:

1. The `["*"]` default is the second half of the cross-origin exposure described above. Defaulting to the same localhost origin set the SDK uses would fix the compose case without opening the server to arbitrary web pages.

2. If someone sets **only** `FASTMCP_ALLOWED_HOSTS` (the natural thing to do when chasing a 421), rebinding protection switches on with an *empty* origin allowlist. The SDK's `_validate_origin` returns `True` only when `Origin` is absent, so every browser request is rejected — while CORS still advertises `*`. Verified with `FASTMCP_ALLOWED_HOSTS=127.0.0.1:8031` and nothing else:

```
OPTIONS /mcp -> 200, Access-Control-Allow-Origin: *
POST    /mcp -> 403 Invalid Origin header
```

A browser client gets a green preflight and then a hard failure, which is a rough thing to debug. Deriving the CORS origins from the same source of truth as `allowed_origins` (and defaulting both to the localhost set) makes the two impossible to disagree.

src/kindly_web_search_mcp_server/server.py
```diff
@@ -92,8 +103,11 @@ def _build_arg_parser() -> argparse.ArgumentParser:
 
 
 def _resolve_transport(raw: str | None) -> Transport:
-    if raw in ("stdio", "sse", "streamable-http"):
-        return raw
+    resolved_transport = raw or os.environ.get("FASTMCP_TRANSPORT", "stdio")
+    if resolved_transport == "http":
+        return "streamable-http"
+    if resolved_transport in ("stdio", "sse", "streamable-http"):
+        return resolved_transport
```
shelpuk: Two small things here, neither blocking:

- `--transport` still has `choices=("stdio", "sse", "streamable-http")`, so `http` is accepted from the env var but rejected from the CLI flag. Adding `"http"` to the choices would make the two surfaces consistent.
- An invalid `FASTMCP_TRANSPORT` silently falls back to stdio. That matches the previous behaviour for the CLI value, so I'm not asking you to change it — but for an env var in a compose file, a typo now means the container comes up in stdio and exits rather than serving HTTP, with no explanation. A `LOGGER.warning` on the fallback would save someone an afternoon.

This function is also the natural place for the tests I mentioned in the summary: env var honoured, `http` aliasing, CLI flag winning over the env var, and the invalid-value fallback.

src/kindly_web_search_mcp_server/server.py
```diff
@@ -176,11 +190,28 @@ def main(argv: list[str] | None = None) -> None:
...
+            allow_origins=allowed_origins if allowed_origins else ["*"],
+            allow_credentials=False,
+            allow_methods=["*"],
+            allow_headers=["*"],
+            expose_headers=["mcp-session-id", "mcp-protocol-version"],
+        )
+        uvicorn.run(app, host=host, port=port)
+    else:
+        try:
+            mcp.run(transport=transport, mount_path=args.mount_path)
```
shelpuk, line 314: `args.mount_path` is now only reachable for stdio, where it does nothing — the HTTP/SSE branch above never passes it. So `--mount-path` has become a no-op CLI option. Fixing the `sse_app()` selection as suggested above (`mcp.sse_app(args.mount_path)`) restores it.

src/kindly_web_search_mcp_server/server.py
```diff
@@ -9,6 +9,9 @@
 from typing import Literal
 
 from mcp.server.fastmcp import FastMCP
+from mcp.server.transport_security import TransportSecuritySettings
+import uvicorn
+from starlette.middleware.cors import CORSMiddleware
```
shelpuk, line 15: `uvicorn` and `starlette` are now imported directly, but neither is declared in `pyproject.toml` — they're currently satisfied transitively through `mcp`. That works today and will keep working right up until it doesn't.

This repo has `tests/test_dependency_constraints.py` specifically because an undeclared/unbounded dependency has bitten us before. Please add both to `[project].dependencies`.

README.md
```diff
@@ -760,6 +760,36 @@ docker run --rm -p 8000:8000 \
...
+      - "8000:8000"
+    environment:
+      - SEARXNG_BASE_URL=...
+      - GITHUB_TOKEN=${KINDLY_GITHUB_TOKEN:-}
+      - FASTMCP_HOST=0.0.0.0
+      - FASTMCP_PORT=8000
+      - FASTMCP_TRANSPORT=http
+```
+
+`FASTMCP_TRANSPORT` can be `streamable-http` (or simply `http`) for Streamable HTTP, `sse` for SSE.
```
shelpuk, line 821: This documents `sse` as a supported value, but with the `hasattr` selection above it doesn't work — `FASTMCP_TRANSPORT=sse` serves `/mcp` and `/sse` 404s. Once the transport selection is fixed this line is accurate as written.

README.md
```diff
@@ -760,6 +760,36 @@ docker run --rm -p 8000:8000 \
...
+
+Run with:
+
+```bash
+docker compose up -d
+```
+
+Container will be built at the first run. To rebuild it, append `--build` to the command above.
```
shelpuk, line 831: Thanks for writing this up — a compose section is a genuinely useful addition and I'd like to keep it.

Two gaps to close before merge:

- `FASTMCP_ALLOWED_HOSTS` and `FASTMCP_ALLOWED_ORIGINS` are the headline feature of this PR and aren't documented anywhere. They belong here, with the `host:*` wildcard syntax spelled out (a bare `*` does not match — see my comment on the `transport_security` block).
- The Docker section immediately above this one ends with a warning that remote HTTP is unauthenticated and unencrypted and shouldn't be exposed publicly. This example publishes `8000:8000` on all host interfaces, so it needs the same warning — more so, given it's the copy-paste-friendly one.

shelpuk:
@niello — heads-up: rather than leave you to carry the security and SSE bits alone, I've pushed three commits to this branch (thanks for leaving "allow edits by maintainers" on). Your four commits are untouched; the diagnosis and the overall shape are yours, I only closed the review points. Shout if you'd rather have done any of it differently.

**Blocking items**

1. **SSE.** Selection now branches on the resolved transport instead of `hasattr`, so `--sse` serves `/sse` again. `mcp.sse_app(args.mount_path)` also brings `--mount-path` back from being a no-op.
2. **Default security.** DNS rebinding protection is now unconditional; an unset allowlist falls back to the loopback set instead of to "no restriction". The `["*"]` CORS fallback is gone.
3. **Half-configured state.** `FASTMCP_ALLOWED_HOSTS` and `FASTMCP_ALLOWED_ORIGINS` now default independently, so setting only the hosts keeps loopback origins working. CORS is compiled from the same origin list the SDK validates against, so the two can't disagree.

**One thing worth knowing** — a detail I got wrong in my own review. The SDK's loopback default is applied only when the *constructor* `host` is a loopback address (`FastMCP.__init__`), and this server sets `mcp.settings.host` later from `FASTMCP_HOST`. So passing `transport_security=None` would have been fragile. The branch now states the loopback allowlist explicitly instead.

Also: Starlette's `CORSMiddleware` matches origins by **exact string** — it doesn't understand the SDK's `host:*` port wildcard at all. Passing `http://localhost:*` to it would have quietly matched nothing. The CORS policy is now built as a regex from the same list, which is what actually makes the two agree.

**Non-blocking items**

- `--transport http` is accepted now, matching what the env var already took.
- An unrecognised `FASTMCP_TRANSPORT` logs a warning before falling back to stdio.
- `start-mcp-server` no longer injects `--stdio` when `FASTMCP_TRANSPORT` is set — that was the one entry point ignoring your variable. An explicit flag still wins.
- `starlette` and `uvicorn` are declared in `pyproject.toml`, with a guard in `tests/test_dependency_constraints.py`.
- README gains a section on both allowlist variables: the `host:*` wildcard, that a bare `*` matches nothing, and that setting a variable **replaces** the loopback defaults rather than adding to them. The Compose example sets `FASTMCP_ALLOWED_HOSTS` (it doesn't work without it) and now carries the same "unauthenticated and unencrypted" warning as the `docker run` section.

**Verification**

24 new tests in `tests/test_server_transport.py` plus 3 in `tests/test_uvx_cli.py`. Full suite: **11 failed, 169 passed** — the same 11 pre-existing Windows/nodriver failures as before, no new ones.

Checked against a live server on both configurations:

| Check | Result |
|---|---|
| `POST /mcp`, `Origin: https://evil.example`, nothing configured | `403` |
| `POST /mcp`, `Host: kindly-web-search-mcp:8077`, nothing configured | `421` |
| preflight from a foreign origin | no `Access-Control-Allow-Origin` |
| preflight + `POST` from `http://localhost:3000` | `200`, origin echoed |
| `POST /mcp` by service name, with `FASTMCP_ALLOWED_HOSTS` set | `200` |
| loopback origin, with only `FASTMCP_ALLOWED_HOSTS` set | `200` (was the 403 half-state) |
| foreign origin, with only `FASTMCP_ALLOWED_HOSTS` set | `403` |
| `--sse`: `GET /sse` | `200`, stream held open |
| `--sse`: `GET /mcp` | `404` |

One thing I did **not** add: a deliberate "allow any host" opt-out. Listing the hostnames you actually use covers the Compose and reverse-proxy cases, and an off switch on an unauthenticated server that can fetch arbitrary URLs didn't seem worth the surface. Easy to add later if someone hits a case that needs it.

Merging this now. Thanks for the report and the fix — the 421/405 diagnosis was spot on and the compose docs are a genuinely useful addition.

niello:
Thank you for fixing the remaining issues. This is well outside my primary area of interest, so I indeed missed a lot of things, and fixing them by myself would have taken a lot of time and several iterations. I really appreciate your effort spent on the review. There is definitely something for me to learn from it.

PR #51, opened 2026-08-30
docs: add the test suite design document

shelpuk:

No product behaviour changes here — this PR adds one document and no code. What it buys is confidence in future releases.

The short version: **our test suite has been quietly broken, and we didn't notice because we'd normalised it.** Every run reports "11 failed" and everyone reads past it. Those 11 aren't a flaky browser — they're tests that stopped matching the code and now crash before they check anything.

The part that matters commercially: they cluster in the page-fetching engine that turns a URL into readable content — browser startup, timeouts, process cleanup. Two specific things there are untested: the worker's start-and-cleanup path, and the browser pool, which has no tests at all. Those are what would leave stray Chromium processes eating a server's memory. Most of the surrounding code *is* tested and passing, so this is a targeted hole, not a collapse.

There is also a security finding the document acts on: **diagnostics are not sanitised, and they are returned to the client.** A URL containing a credential comes back over the wire verbatim. Separately, the server will fetch any URL it's given from wherever it runs, and we have never written down whether that should be restricted. Neither is introduced here — both are existing conditions this document makes visible and testable.

This is the plan to fix all of it: what to test, at which level, with which tools, and in what order. First job is making the suite honest again; **enforcement lands immediately after**, so a regression becomes un-mergeable before the rest of the work is built; then the risky areas get real coverage. We currently have no CI at all.

**Cost.** Engineering time, plus one small recurring item: the nightly Serper canary spends the `SERPER_API_KEY` at one query per night. Everything else is mocked and free.

## What's in the document

`.system_design/TEST_SUITE.md`. This also establishes `.system_design/` in the repo.

A four-layer test model applied to this system specifically:

- **L1 — component/property.** URL parsers, diagnostics redaction, env resolvers, launch-argument builders, Markdown transforms. Property-based where it earns it: *no URL may match two parsers* (the resolver is a first-match chain, so overlap is a silent mis-route).
- **L2 — contract.** Keeps the three good ones. Adds the MCP tool schema, the `KINDLY_DIAG` parent↔worker frame format, and response-model invariants.
- **L3 — subsystem.** Real socket, real child process, real Chromium — split into portable tests that run on Windows and Linux, and Chromium tests that need a container.
- **L4 — product.** A deterministic MCP session driven from an installed wheel on every PR, plus gated live canaries nightly.

Plus security testing (§7), a risk-to-test matrix covering every subsystem (§9), CI and enforcement (§10.3), and the production seams the design depends on (§11).

## Ordering

Workstreams run: **A** restore green → **B** enforce it (minimal gate + branch protection) → **C** production seams → **D** new coverage → **E** async migration. B sits second deliberately: the original failure was not that tests broke, it was that a red suite was never enforced and became normal.

## Review history

Several review rounds, all reflected in the head commit. The findings that changed the design most:

- **The baseline is platform-dependent** — eight stale callers, one of which skips on Windows — and `scrape/` is *not* wholly untested. The document names the specific uncovered functions instead.
- **`_split_worker_diagnostics` is dead code**, so the contract test was retargeted at the live streaming path.
- **`outputSchema` is `null` for both tools** — they're annotated `-> dict`, so FastMCP derives no result schema. §4.2 pins that fact rather than asserting a shape that doesn't exist.
- **The stale sandbox tests assert two different things** — flag resolution *and* retry orchestration. They're split across layers rather than relocated wholesale.
- **`fetch_html_via_nodriver` hardcodes its child command**, so §11 schedules a private `_run_worker_command` seam.
- **A `.pdf` URL is not automatically network-free**: `parse_arxiv_url` accepts `https://arxiv.org/pdf/2401.12345.pdf`, so the wheel smoke test pins a non-specialized host.
- **`ci-required` would not have failed on a failing dependency** — GitHub skips `needs` jobs rather than failing them, so the aggregator uses `if: always()` and asserts every result is `success`.
- **The coverage baseline was a floor, not a ratchet**, and exact equality needed a *pinned environment*, not just a fixed job set. It now runs in a pinned lane with compatibility axes excluded.
- **The production Docker image cannot be the test image.** It installs Chromium from a rolling archive and bakes `src/` in, so it either floats or tests stale code. A separate pinned test-runtime image takes the PR's wheel at run time.

Four claims of my own were corrected by measurement rather than argument: `pytest -m <unknown>` exits **5**, not 0; `pytest_collection_modifyitems` runs before mark deselection and never fires; `create_autospec` on `asyncio.subprocess.Process` declares no `stdout`/`stderr`/`pid`, so it would have accepted the very fake that broke; and `@runtime_checkable` returns `True` for a wrong method signature, so the Protocol needs a static type check behind it.

## Decisions recorded with their reasoning

- **A saved-HTML corpus instead of live extraction tests.** Pages change, so live assertions either weaken to nothing or flake.
- **Real child processes *and* narrow doubles at L3.** A fixture child can't assert which Chromium flags were chosen; a double can't assert a process tree died.
- **Configuration, not a production hook, for cross-process stubbing.** An injection point in shipped code is a permanent risk on an unauthenticated server.
- **No absolute coverage target**, but an enforced diff-coverage gate and a ratcheted baseline in a pinned lane.
- `mutmut` needs `fork()`, confirmed against its docs, so mutation runs are Linux-CI only and never in the local Windows loop.

## Four open decisions (§13)

1. **Outbound request policy** — is fetching private-network addresses intentional? Plausibly yes, given self-hosted SearXNG; a blanket block would break those users. Undefined today, on an unauthenticated server.
2. **Tool result schemas** — annotate the tools with their response models so clients get an `outputSchema`, or keep `-> dict`? An API change with client impact.
3. **Owners** in the risk matrix.
4. **Per-module coverage floors** — deferred until workstream A produces a measured baseline.

## Verification

Every factual claim was checked against the working tree. Baseline re-run: **11 failed, 169 passed, 3 skipped** on Windows / CPython 3.13 / mcp 1.25.0. The equivalent POSIX figure (12) is labelled as read-from-source, not executed — measuring it is workstream A step 1.

## Known gaps

`.system_design/SYSTEM_DESIGN.md` still doesn't exist, so this document describes testing a system whose design was never written down; §14. Incidentally noted and not touched: `uv.lock` is referenced in `pyproject.toml` and `tests/test_dependency_constraints.py` but does not exist, and `requirements.txt` omits `trafilatura`.

https://claude.ai/code/session_013kB3d2mEtUkkANj92V8Nt7

### DIFF
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -1502,33 +1502,59 @@ Stated explicitly so the omission is not mistaken for an oversight.
 
 ---
 
-## 13. Open decisions
+## 13. Decision records and external prerequisites
 
-Four decisions cannot be settled by this document. Each is recorded here in full:
-what it decides, why it is open, what it blocks, and what an answer must contain.
+Not everything here is open, and not everything open is here. The four states:
 
-**This section is the authoritative statement of each decision.**
-`TEST_SUITE_IMPLEMENTATION_PLAN.md` §3 tracks the same items as external
-prerequisites X-1…X-6 with owners and blocked steps, but for *scheduling* only. If
-the two disagree about what is being decided, this section is right and the plan
-is stale.
+| State | Items |
+|---|---|
+| **Externally unresolved** — nobody has decided | §13.1 (X-1), §13.3 (X-6) |
+| **Provisionally resolved** — settled for this implementation, revisitable | §13.2 |
+| **Scheduled** — deferred deliberately, with a step that discharges it | §13.4 |
+| **Deferred with a reason** — not scheduled, and why | §13.5 |
+
+**Operational prerequisites live only in the plan.** X-2 (repository admin), X-3
+(container registry), X-4 (live-query budget) and X-5 (observational coverage
+owner) are provisioning and ownership tasks, not design decisions, so they are
+tracked in `TEST_SUITE_IMPLEMENTATION_PLAN.md` §3 and are deliberately absent
+here.
+
+**For the items that are here, this section is authoritative.** The plan's §3
+records the same decisions for *scheduling* — owner, blocked steps — but not what
+is being decided. If the two disagree about substance, this section is right and
+the plan is stale.
+
+**Review points we decide not to act on are recorded inline** as `> **Deferred:**`
+blocks at the place they apply, naming what was raised and why it was not taken.
+A judgement that lives only in a review thread is invisible to the next reader.
 
 ### 13.1 Outbound request policy — X-1
 
-**The largest open item, and it is three decisions rather than one.** Answering
+**The largest open item, and it is four decisions rather than one.** Answering
 only the first leaves the browser path unenforced, which is why this keeps
 resurfacing in review.
 
-1. **Is fetching private-network addresses intentional?** It plausibly is —
-   self-hosted SearXNG and internal documentation are exactly what this server is
-   pointed at, and a blanket RFC1918 block would break those users.
-2. **If restricted, how are Chromium's own connections enforced?** `httpx`
+**Decisions 1 and 2 are independent dimensions, not one switch.** Treating them as
+a single allow/restrict choice cannot express what may well be the right answer —
+*permit private HTTP(S) so self-hosted SearXNG and intranet docs keep working,
+while refusing `file:`, `data:`, `ftp:` and `chrome:` outright*. Under a single
+switch, "allow" would delete the enforcement steps that a scheme restriction still
+needs.
+
+1. **Which URL schemes are fetchable?** `http` and `https` are the only ones the
+   loaders are built for; `file:`, `data:`, `ftp:` and `chrome:` are reachable
+   today and each has a different failure mode.
+2. **Are private-network addresses fetchable?** Plausibly yes — self-hosted
+   SearXNG and internal documentation are exactly what this server is pointed at,
+   and a blanket RFC1918 block would break those users.
+3. **If *either* dimension is restricted, how are Chromium's own connections
+   enforced?** `httpx`
    resolves in Python where the address is observable. Chromium resolves and
    connects *inside the browser process*, so a Python-side seam proves nothing
    about what it contacted. Candidates: a **local policy-enforcing proxy** that
    resolves and decides; a **host-resolver rule paired with per-request
    interception**; or another mechanism mediating every browser request.
-3. **What happens to an upstream proxy?** `KINDLY_CHROME_PROXY` is passed to
+4. **What happens to an upstream proxy?** `KINDLY_CHROME_PROXY` is passed to
    Chromium verbatim (`nodriver_worker.py`, `_build_chromium_launch_args`). With
    an HTTP `CONNECT` proxy the **proxy** resolves the hostname and the server never
    sees the destination address, so host-side validation is bypassed by
@@ -1536,12 +1562,25 @@ resurfacing in review.
    trusted boundary whose limit is documented, or chain it through something that
    enforces.
 
-Decisions 2 and 3 are independent. Disallowing or trusting the proxy says nothing
-about Chromium with no proxy configured.
+Decisions 3 and 4 are independent of each other too. Disallowing or trusting the
+proxy says nothing about Chromium with no proxy configured.
+
+> **Deferred: enumerating further policy dimensions.** Scheme and address class are
+> separated above because both are concretely reachable today and pull in opposite
+> directions — the plausible answer permits private HTTP(S) while refusing
+> `file:`/`data:`. Other dimensions could be carved out in the same way: redirect
+> depth, non-standard ports, request method, response size. They are **not**
+> enumerated here, because each one that is speculative adds a branch E9-0 must
+> materialise and a column the answer must fill, for a restriction nobody has asked
+> for. If X-1's author needs one, adding it is a paragraph here and a case in
+> §7.2 — the structure supports it; the guess does not belong in it.
 
 **An answer must:**
 
-- State the policy, and for a restrictive policy name the mechanisms for 2 and 3.
+- State the policy **per dimension** — schemes and address classes separately —
+  and, if either restricts anything, name the mechanisms for 3 and 4. Enforcement
+  is needed whenever *any* dimension restricts *anything*; only a policy permitting
+  every listed scheme **and** every listed address class removes the need for it.
 - **Mediate every request the browser makes, not the navigation URL.** A permitted
   public page can reach a private address through an image, script, frame,
   stylesheet or `fetch`. An architecture that passes top-level navigation and
@@ -1551,8 +1590,9 @@ about Chromium with no proxy configured.
 - Re-plan the implementation branch. A restrictive policy is not one change: it
   spans a shared policy function, `httpx` enforcement across redirect hops,
   Chromium enforcement, and deterministic rebinding and proxy tests. The plan's
-  **E9-0** exists to carry that re-planning and must land before any security step
-  is authored.
+  **E9-0** exists to carry that re-planning and must land before any
+  **outbound-policy** step is authored — the plan's E3-6 and E9-2…E9-7. It does
+  not gate E9-1, the diagnostics sanitizer, which is independent of this decision.
 
 **Blocks:** all outbound-security work — the plan's E9-0, and through it E3-6 and
 E9-2…E9-7. Nothing else in the plan waits on it.
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -545,19 +545,36 @@ supplies, from wherever the server runs**. Nothing currently restricts:
 
 - Non-HTTP schemes — `file:`, `data:`, `ftp:`, `chrome:`.
 - Loopback, RFC1918, link-local, IPv6 unique-local, and cloud metadata addresses
-  (`169.254.169.254`).
+  (`169.254.169.254`), including IPv4-mapped IPv6 forms.
 - Redirects that begin public and land private.
 - DNS rebinding between validation and connection (TOCTOU).
+- **A hostname resolving to several addresses at once**, where validation may
+  inspect a public record while the connector selects a private one. This needs no
+  timing and is not rebinding.
+- **Subresources fetched by a permitted page.** Chromium loads images, scripts,
+  frames, stylesheets and `fetch()` targets on its own account. Restricting the
+  navigation URL while the loaded page reaches `169.254.169.254` is still SSRF, so
+  any enforcement mechanism has to mediate *every* request the browser makes, not
+  the one it was given.
 - Proxy interaction, where `KINDLY_CHROME_PROXY` may reach networks the host
-  cannot.
+  cannot — and where, with an HTTP `CONNECT` proxy, the **proxy** resolves the
+  hostname and the server never sees the destination address.
 
 **The policy is a product decision** (§13). Once stated, tests follow at three
 boundaries and the shape is the same either way:
 
-1. **URL validation** — L1, table-driven over scheme and address class.
+1. **URL validation** — L1, table-driven over scheme, address class, and
+   multi-record answers including mixed public/private and dual-family results.
 2. **Redirect handling** — L3, a local server issuing a public→private redirect.
 3. **Connection** — L3, a hostname whose resolution changes between validation
    and connect.
+4. **Browser-initiated requests** — L3, a permitted public page attempting a
+   private subresource by each route Chromium supports.
+
+The `httpx` and Chromium paths need separate mechanisms: `httpx` resolves in
+Python where the address is observable, while Chromium resolves and connects
+inside the browser process. Selecting the browser mechanism is part of §13.1's
+decision, not an implementation detail beneath it.
 
 ---
 
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -1504,29 +1504,103 @@ Stated explicitly so the omission is not mistaken for an oversight.
 
 ## 13. Open decisions
 
-Four items need an answer from the maintainer; none can be settled by this
-document.
+Four decisions cannot be settled by this document. Each is recorded here in full:
+what it decides, why it is open, what it blocks, and what an answer must contain.
 
-1. **Outbound request policy (§7.2).** Is fetching private-network addresses
-   intentional? It plausibly is — self-hosted SearXNG and internal documentation
-   are exactly what this server is pointed at, and a blanket RFC1918 block would
-   break those users. The tests in §7.2 apply either way; what is not acceptable
-   is leaving it undefined on an unauthenticated server.
-2. **Tool result schemas (§11.3).** Annotate the tools with their response models
-   so clients receive an `outputSchema`, or keep `-> dict` and pin its absence?
-   This is an API change with client impact.
-3. **Owners in the risk matrix (§9).** Assigning maintainers is not this
-   document's call; the column is otherwise complete.
-4. **Per-module coverage minimums.** Diff coverage and the committed baseline are adopted
-   (§10.4); fixed per-module floors are not, because no coverage measurement
-   exists in this repo yet and any number chosen now would be invented. Revisit
-   once workstream A produces a measured baseline.
+**This section is the authoritative statement of each decision.**
+`TEST_SUITE_IMPLEMENTATION_PLAN.md` §3 tracks the same items as external
+prerequisites X-1…X-6 with owners and blocked steps, but for *scheduling* only. If
+the two disagree about what is being decided, this section is right and the plan
+is stale.
 
-Deferred with a reason: **versioning the `KINDLY_DIAG` frame format.** There is
-no version field today, so adding one is a wire-format change, not a test
-decision, and the two endpoints ship in the same wheel and cannot disagree by
-version in practice. If a version field is ever added, §4.3 gains an
-unknown-version case; until then there is nothing to test.
+### 13.1 Outbound request policy — X-1
+
+**The largest open item, and it is three decisions rather than one.** Answering
+only the first leaves the browser path unenforced, which is why this keeps
+resurfacing in review.
+
+1. **Is fetching private-network addresses intentional?** It plausibly is —
+   self-hosted SearXNG and internal documentation are exactly what this server is
+   pointed at, and a blanket RFC1918 block would break those users.
+2. **If restricted, how are Chromium's own connections enforced?** `httpx`
+   resolves in Python where the address is observable. Chromium resolves and
+   connects *inside the browser process*, so a Python-side seam proves nothing
+   about what it contacted. Candidates: a **local policy-enforcing proxy** that
+   resolves and decides; a **host-resolver rule paired with per-request
+   interception**; or another mechanism mediating every browser request.
+3. **What happens to an upstream proxy?** `KINDLY_CHROME_PROXY` is passed to
+   Chromium verbatim (`nodriver_worker.py`, `_build_chromium_launch_args`). With
+   an HTTP `CONNECT` proxy the **proxy** resolves the hostname and the server never
+   sees the destination address, so host-side validation is bypassed by
+   configuration alone. Prohibit it under a restrictive policy, treat it as a
+   trusted boundary whose limit is documented, or chain it through something that
+   enforces.
+
+Decisions 2 and 3 are independent. Disallowing or trusting the proxy says nothing
+about Chromium with no proxy configured.
+
+**An answer must:**
+
+- State the policy, and for a restrictive policy name the mechanisms for 2 and 3.
+- **Mediate every request the browser makes, not the navigation URL.** A permitted
+  public page can reach a private address through an image, script, frame,
+  stylesheet or `fetch`. An architecture that passes top-level navigation and
+  cannot constrain subresources does not satisfy this decision (§7.2).
+- Say whether *every* address a hostname resolves to must be permitted, or whether
+  the connection is pinned to one validated address (§7.2).
+- Re-plan the implementation branch. A restrictive policy is not one change: it
+  spans a shared policy function, `httpx` enforcement across redirect hops,
+  Chromium enforcement, and deterministic rebinding and proxy tests. The plan's
+  **E9-0** exists to carry that re-planning and must land before any security step
+  is authored.
+
+**Blocks:** all outbound-security work — the plan's E9-0, and through it E3-6 and
+E9-2…E9-7. Nothing else in the plan waits on it.
+
+**Not deciding is itself a position**, and a poor one: the server is
+unauthenticated and `get_content` fetches whatever URL it is given.
+
+### 13.2 Tool result schemas — no external prerequisite
+
+Annotate `web_search` and `get_content` with their response models so clients
+receive an `outputSchema`, or keep `-> dict` and pin its absence? An API change
+with client impact either way.
+
+**Resolved for this implementation, provisionally:** `outputSchema is None` is
+**normative** and §4.2's golden test pins it. That keeps the contract test honest
+about what clients receive today. Revisiting is a new step, not a pending decision
+blocking anything — which is why this has no X-number.
+
+### 13.3 Owners in the risk matrix — X-6
+
+§9's Owner column is blank except where X-4 (live-alert) and X-5 (observational
+coverage) name someone. Assigning the rest is not this document's call.
+
+**An answer must** name an owner per row, and be **written into §9** — the plan's
+**E13-2** does that, because an answer living only in a ticket leaves the
+authoritative design showing an empty column.
+
+**Blocks:** completion (the plan's E13-1) via E13-2. It blocks no implementation
+work.
+
+### 13.4 Per-module coverage minimums — no external prerequisite
+
+Diff coverage and the committed baseline are adopted (§10.4). Fixed per-module
+floors are not, because no coverage measurement exists in this repository yet and
+any number chosen now would be invented.
+
+**Discharged by the plan's E10-11**, once E10-3 produces the first real
+measurement — not "workstream A", which was the earlier and wrong reference.
+**"Do not adopt floors" is a valid outcome** provided the measured figures are
+recorded here; leaving this open indefinitely is not.
+
+### 13.5 Deferred, with a reason
+
+**Versioning the `KINDLY_DIAG` frame format.** There is no version field today, so
+adding one is a wire-format change rather than a test decision, and the two
+endpoints ship in the same wheel and cannot disagree by version in practice. If a
+version field is ever added, §4.3 gains an unknown-version case; until then there
+is nothing to test.
 
 ---
 
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -774,8 +774,19 @@ these controls need exactly one pinned environment.
 The job is deliberately **self-contained — it consumes no artefacts from other
 jobs**:
 
-1. Installs from `requirements-ratchet.txt` on Python 3.13, from a source
-   checkout.
+1. Installs from `requirements-ratchet.txt` on Python 3.13, then makes the project
+   itself importable — the lockfile deliberately excludes it, so an explicit step
+   is required and the two commands belong together:
+
+   ```bash
+   pip install -r requirements-ratchet.txt
+   pip install --no-deps -e .
+   ```
+
+   The job then asserts `kindly_web_search_mcp_server.__file__` resolves under the
+   checkout's `src/`, which is the mirror image of §6.1's assertion and catches the
+   opposite mistake: measuring an installed wheel when the lane is supposed to
+   measure the working tree.
 2. Runs the hermetic selection itself
    (`--ignore=tests/package -m "not live and not subsystem and not chromium and not package"`)
    under `coverage run`.
@@ -1007,9 +1018,10 @@ check** under branch protection on `main`, with `fast`, `fast-extras`,
 
 ### 10.4 Coverage
 
-No absolute percentage target: a percentage rewards executing lines and can be
-satisfied by deleting tests. Three controls, each aimed at a failure the others
-cannot see.
+No absolute percentage target: a percentage rewards executing lines, and it rises
+when you delete uncovered production code — a refactor that removes a dead branch
+improves the number without improving the testing. Three controls instead, each
+aimed at a failure the others cannot see.
 
 **1. Every production module must appear in the report.** This is the control that
 would actually have caught this project's worst gap, and the first draft of this
@@ -1021,8 +1033,11 @@ no test ever imports is absent from the report entirely — so
 and drag nothing down. It would be invisible rather than visibly at zero. Setting
 `source_pkgs` is what makes coverage.py report never-executed files.
 
+Three configuration files, because the jobs genuinely need different behaviour and
+an earlier draft described settings that were not in the file it displayed:
+
 ```ini
-# .coveragerc
+# .coveragerc — base, used by the gating lane's control 1 view
 [run]
 source_pkgs = kindly_web_search_mcp_server
 branch = true
@@ -1034,36 +1049,105 @@ source =
     */site-packages/kindly_web_search_mcp_server
 ```
 
-The `[paths]` mapping is not optional here. The lane installs a wheel, so coverage
-records paths under `site-packages`, while `git diff` produces
-`src/kindly_web_search_mcp_server/...`. Unmapped, the two never join: the totals
-double-count and `diff-cover` finds no coverage for any changed line and reports
-nothing — passing silently. `relative_files = true` is what keeps the mapping
-working across a Windows and Linux combine.
+```ini
+# .coveragerc-gate — controls 2 and 3; base plus the declared exemptions
+[run]
+source_pkgs = kindly_web_search_mcp_server
+branch = true
+relative_files = true
+omit =
+    */scrape/nodriver_worker.py
+    */scrape/chromium_pool.py
+    */scrape/worker_runner.py
 
-Two assertions enforce it:
+[paths]
+source =
+    src/kindly_web_search_mcp_server
+    */site-packages/kindly_web_search_mcp_server
+```
 
-- Every `.py` file under `src/kindly_web_search_mcp_server/` appears in the
-  report, including files at zero. A new module that nothing imports fails the
-  build rather than vanishing.
-- **If the diff contains at least one changed line that the coverage report
-  represents as a statement**, `diff-cover` may not report zero analyzed lines.
-  Zero in that case means the path mapping broke, not that the change was safe.
-  The condition is necessary rather than pedantic: a PR editing only a literal on
-  a continuation line of a multi-line call legitimately has no represented
-  statement lines, and an unconditional guard would misdiagnose that as broken
-  wiring.
+```ini
+# .coveragerc-subprocess — observational L3 jobs only
+[run]
+source_pkgs = kindly_web_search_mcp_server
+branch = true
+relative_files = true
+parallel = true
+patch = subprocess
+```
 
-**One coverage product, from the hermetic lane.** Every control measures the
-`coverage` job's own pinned run of the L1/L2 selection. Nothing is combined across
-jobs.
+Each job selects one by **absolute `COVERAGE_RCFILE`**, never by discovery.
+Discovery walks up from the working directory, and §6.1 deliberately runs the
+package tests from outside the checkout, where the file would not be found.
+`COVERAGE_FILE` is likewise set to an absolute path.
 
-That is a deliberate reduction in what the coverage controls claim, and it costs
-something: changes to code only reachable through `subsystem` or `chromium` —
-the worker lifecycle, `ChromiumPool` — are **not** diff-gated. What covers them
-instead is §2.1's allocation rule enforced in review, the `subsystem` and
-`chromium` jobs themselves being required, and control 1 below, which still fails
-if such a module has no tests at all.
+`patch = subprocess` implies `parallel = true`, so those jobs write suffixed data
+files and **must run a local `coverage combine` before reporting** — reporting does
+not implicitly combine them. That combine is local to the job; it is not the
+cross-job fan-in removed above.
+
+The `[paths]` mapping stays in the gating configs even though that lane runs from a
+source checkout, because §6.1's package tests do install a wheel and their
+observational report needs the mapping to line up with the tree.
+
+The `[paths]` mapping matters wherever a wheel is involved. The gating lane runs
+from a source checkout, so its own paths already match `git diff`; §6.1's package
+tests install a wheel and record paths under `site-packages`, and unmapped those
+never join `src/kindly_web_search_mcp_server/...` — the totals double-count and
+`diff-cover` finds no coverage for any changed line, passing silently.
+`relative_files = true` keeps paths comparable between jobs and platforms.
+
+Three assertions enforce it. Note that "appears in the report" is **not** one of
+them: under `source_pkgs` every module always appears, so such a check can only
+fail on a broken config and says nothing about whether tests exist. The control
+has to assert coverage, not presence.
+
+- **Classification is total and exclusive.** Every `src/**/*.py` file is either in
+  the gating scope or in `.coveragerc-gate`'s `omit` list — exactly one. A policy
+  test walks the tree and the omit list and fails on any file in both or neither,
+  so a new module cannot slip in unclassified.
+- **Every gating-scope module has non-zero coverage** in the hermetic report, and
+  **every omitted module has non-zero coverage** in the observational L3 report.
+  This is what would have caught `chromium_pool.py`: it is classified L3, the L3
+  report shows it at zero, the build fails until a test exists. Presence never
+  would have.
+- **If the diff contains at least one changed line the report represents as a
+  statement**, `diff-cover` may not report zero analyzed lines — that means the
+  path mapping broke, not that the change was safe. The condition is necessary
+  rather than pedantic: a PR editing only a literal on a continuation line of a
+  multi-line call legitimately has no represented statement lines, and an
+  unconditional guard would misdiagnose that as broken wiring.
+
+**One coverage run, two report views.** Every control measures the `coverage`
+job's own pinned run of the L1/L2 selection; nothing is combined across jobs. But
+that single run is reported twice, and the difference is load-bearing.
+
+`source_pkgs` makes the report **whole-package by definition** — that is the point
+of control 1, and it is also a trap. A module the hermetic lane cannot execute,
+such as `chromium_pool.py`, appears with every statement at zero hits. It is not
+absent; it is present and uncovered. So a scope claim cannot be made by wishing:
+without filtering, `diff-cover` sees those zero-hit statements and counts changed
+ones as uncovered, and adding L3-only statements grows the ratchet's denominator
+while the numerator stands still. Ordinary worker or pool development would fail
+both gates, pushing authors either to write hermetic tests for code that has no
+hermetic seam or to reach for the baseline-reset escape hatch. An earlier draft of
+this section asserted such lines were "not counted for or against" the gate; that
+was false, and this is the correction.
+
+**The gating scope is therefore declared, and the declaration is checked.**
+
+| View | Config | Feeds |
+|---|---|---|
+| whole-package | `.coveragerc` | control 1 |
+| gating scope | `.coveragerc-gate` — the same file plus an `omit` list | controls 2 and 3 |
+
+`omit` names the modules with no hermetic seam. Today that is
+`scrape/nodriver_worker.py`, `scrape/chromium_pool.py`, and `scrape/worker_runner.py`
+— the process-management module §11.2 extracts from `universal_html.py`.
+
+**That extraction is a prerequisite of this control, not a nicety** — see §11.2
+for why file-granularity `omit` forces it, and why the split is the same argument
+§2.1 makes about assertions: the seam belongs in the design, not in a side-table.
 
 The alternative was combining every lane into the diff gate, and it does not
 survive contact with the arithmetic. A required threshold gate must take
@@ -1075,9 +1159,25 @@ flipping the threshold; "rarely" is not a property a required check can be built
 on, and a flaky required check is precisely the normalised-red failure this whole
 document exists to prevent.
 
-The full `subsystem` and `chromium` coverage is still worth producing and reading
-as **observational** reporting — it is how someone notices the worker is thinly
-covered — but it does not gate.
+The full `subsystem` and `chromium` coverage is still produced as
+**observational** reporting. It does not gate — but "observational" cannot mean
+"emitted and forgotten", because this project's founding problem was exactly that:
+a signal nobody read. Unowned reporting decays into the same thing.
+
+So it is specified rather than gestured at:
+
+- Published as an HTML and JSON artefact per run, with `if-no-files-found: error`,
+  so a silently missing report fails the job instead of reading as "nothing to
+  say".
+- Reduced to a **per-module summary in the PR job summary** — module, percent,
+  delta against the previous run on `main`. A number in a downloadable artefact is
+  not a signal; a table on the PR is.
+- Control 1's second assertion already gates the strongest fact this report
+  carries — that an L3-classified module has *some* coverage — so the observational
+  view is for degree, not existence.
+- It has a **named reviewer and a cadence**, assigned when §13's owner decision is
+  resolved. Until then it is explicitly unowned, which is a known gap rather than
+  an oversight.
 
 **2. Diff coverage — the PR gate.** `diff-cover --fail-under=80` over the changed
 lines of each PR.
@@ -1145,7 +1245,7 @@ separate decision with its own runtime cost, not an answer to this gap.
 the hermetic suite is written to `coverage-baseline.json`, committed, and may not
 decrease.
 
-**Scope: L1 and L2 only — `fast` on Linux, Python 3.13, pinned lane.** The
+**Scope: the gating scope only, measured on Linux, Python 3.13, pinned lane.**
 This follows from the single-product decision above rather than adding a rule:
 every control measures the hermetic lane, so the ratchet inherits its determinism.
 Were the nondeterministic lanes included, a single differently-taken branch would
@@ -1319,6 +1419,20 @@ There is no way to point it at a fixture child. Split the function in two:
 `fetch_html_via_nodriver` keeps its current signature and **always builds its own
 command**, calling the builder and passing the result to the runner. Tests reach
 `_run_worker_command` directly with a fixture command.
+
+**Move the runner into its own module, `scrape/worker_runner.py`.** This is not
+tidiness: §10.4 classifies every production file as hermetically testable or not,
+and `omit` works at file granularity. `universal_html.py` today mixes the
+Markdown-suffix probe path — covered by 15 passing hermetic tests — with subprocess
+spawn and stream management that only a real child can exercise. While both live in
+one file, the coverage gate must treat the whole file as one or the other, and both
+choices are wrong: gate it and worker changes fail against a lane with no seam for
+them; exempt it and the probe path loses real gating. The split makes the
+classification a module boundary, which is a design property rather than a
+side-table to maintain.
+
+`_build_worker_command` stays with `fetch_html_via_nodriver`; it is pure and
+hermetically testable, so it belongs on the gated side.
 
 **The command must not become a parameter of the public fetch API.** Adding a
 `command=` argument to `fetch_html_via_nodriver` would turn "execute an arbitrary
--- a/.system_design/TEST_SUITE.md
+++ b/.system_design/TEST_SUITE.md
@@ -538,55 +538,89 @@ state, and tests must cover:
 - Low-entropy secret values, where substring matching yields false positives —
   the policy must say whether it redacts by key name, by pattern, or both.
 
-### 7.2 Outbound request policy — undefined
+### 7.2 Outbound request policy
 
-The server is unauthenticated and `get_content` fetches **any URL the caller
-supplies, from wherever the server runs**. Nothing currently restricts:
+**Decided 2026-08-31.** The server is unauthenticated and `get_content` fetches
+whatever URL a caller supplies, so the policy is stated here rather than left to
+implementation.
 
-- Non-HTTP schemes — `file:`, `data:`, `ftp:`, `chrome:`.
-- Loopback, RFC1918, link-local, IPv6 unique-local, and cloud metadata addresses
-  (`169.254.169.254`), including IPv4-mapped IPv6 forms.
-- Redirects that begin public and land private.
-- DNS rebinding between validation and connection (TOCTOU).
-- **A hostname resolving to several addresses at once**, where validation may
-  inspect a public record while the connector selects a private one. This needs no
-  timing and is not rebinding.
-- **Subresources fetched by a permitted page.** Chromium loads images, scripts,
-  frames, stylesheets and `fetch()` targets on its own account. Restricting the
-  navigation URL while the loaded page reaches `169.254.169.254` is still SSRF, so
-  any enforcement mechanism has to mediate *every* request the browser makes, not
-  the one it was given.
-- Proxy interaction, where `KINDLY_CHROME_PROXY` may reach networks the host
-  cannot — and where, with an HTTP `CONNECT` proxy, the **proxy** resolves the
-  hostname and the server never sees the destination address.
+**Scheme.** `http` and `https` only. `file:`, `data:`, `ftp:` and `chrome:` are
+refused. They are reachable today by accident — no loader handles them — and each
+has its own failure mode.
 
-**The policy is a product decision** (§13). Once stated, tests follow at **five**
-boundaries, and they exist whatever the policy says:
+**Address, by provenance.** The two are different code paths and hold different
+policies:
+
+| Provenance | Policy |
+|---|---|
+| **Externally supplied** — `get_content`'s `url`, and the links `web_search` fans out to | **Public addresses only.** Loopback, RFC1918, link-local, IPv6 unique-local, cloud metadata (`169.254.169.254`) and their IPv4-mapped IPv6 forms are refused. |
+| **Operator-configured** — `SEARXNG_BASE_URL`, `KINDLY_CHROME_PROXY` | **Unrestricted.** The operator chose these; reaching a private address here is the intended behaviour and nothing an attacker influences. |
+
+Restricting the externally-supplied path does **not** break a self-hosted SearXNG
+deployment, because that runs through `search_searxng` against an
+operator-configured URL rather than through the resolver.
+
+**A hostname resolving to several addresses** must have *every* returned record
+permitted. Pinning to one validated address would be stronger, but the connectors
+do not expose that control; requiring all records is enforceable with what exists.
+
+**Enforcement mechanism, by stack.** These differ because the stacks differ:
+
+- **`httpx`** — enforced. Resolution happens in Python, so the address is
+  observable and the check runs at validation, at every redirect hop, and against
+  the resolved address set. E3-6's seam is what makes this possible.
+- **Chromium** — **pre-navigation only**. The scheme and the resolved address of
+  the URL handed to the browser are checked before navigation. Chromium then
+  resolves and connects inside its own process, so two cases are **not** enforced:
+
+  1. **DNS rebinding** — the browser re-resolves, and may reach an address the
+     check never saw.
+  2. **Subresources** — a permitted public page can request a private address
+     through an image, script, frame, stylesheet or `fetch`, and the server never
+     sees that URL.
+
+**These two gaps are accepted, not overlooked.** Closing them needs a local
+policy-enforcing proxy that mediates every browser request — a component the
+project would own and maintain. That was judged disproportionate for a server
+whose realistic exposure is a caller naming a private address directly, which
+pre-navigation checking does stop. §7.3 records the trigger for revisiting.
+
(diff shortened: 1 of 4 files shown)


shelpuk:
Revision 2 pushed (`3157950`). The review was largely right, and I verified each factual claim against the tree before acting on it rather than taking them on trust. Four confirmed, and they were load-bearing:

**Confirmed and applied**

- **Eight stale `_fetch_html` callers, not seven.** The eighth skips on Windows because `os.geteuid` doesn't exist there (`test_nodriver_worker_sandbox.py:199`), so POSIX should show 12 failures rather than 11. I've labelled that POSIX number as derived-from-source-not-executed and made measuring it a task, since I can't run Linux here.
- **"`scrape/` has zero effective coverage" was an overstatement.** 15 of 18 tests in `test_universal_html_loader.py` pass and cover the whole Markdown-suffix probe path; `test_worker_launch_args_redaction.py` covers the launch-arg builder. §1.2 now names the actually-uncovered functions. The concurrency failure is in `server.py` — revision 1 contradicted its own table by folding it into `scrape/`.
- **`_split_worker_diagnostics` has no callers anywhere.** I'd designed a contract test around dead code. It now targets the live path (`_read_stderr_stream` → `_consume_stderr_line`), with the fragmentation, CRLF, EOF and split-Unicode cases you listed, and the dead function is flagged for separate removal. I also dropped the claim that this contract would have prevented the incident — it wouldn't.
- **`Diagnostics.emit` applies no redaction at all.** Worse than stated: `entries` is returned to the client in `GetContentResponse.diagnostics`, so an unredacted `https://user:token@host/` egresses over the MCP wire, not just to stderr. Sanitization moved to the serialization boundary with the policy items you listed.

**On the sandbox tests** — you're right that a toy child can't verify flags, but I went further than "repair with autospecced doubles". Those assertions are about *pure functions* (`_build_chromium_launch_args`, `_resolve_sandbox_enabled`), and routing them through an ~480-line browser-launching coroutine is precisely why one signature change disabled all eight. They move down to L1, in the style of the already-passing `test_worker_launch_args_redaction.py`. Real-process tests are added alongside, not instead.

**On coverage** — your correction to my rationale is right and I've said so in the doc: a `TypeError` on a missing kwarg raises *before* the body runs, so coverage would have shown `_fetch_html` going dark. Added a no-regression ratchet and diff coverage.

Also added: outbound request policy (§7.2, absent entirely before), the risk-to-test matrix (§9), corpus governance (§3.3), anti-flake/cleanup rules (§5.4), and the CI split that makes the Windows rationale actually true. All six low-severity accuracy fixes applied.

**Deferred, with reasons in §13** — protocol versioning for `KINDLY_DIAG` (that's a wire-format change, not a test decision, and the endpoints ship in one wheel); per-module coverage floors (no baseline exists yet, so any number would be invented — revisit after §8 measures one); matrix owners (not mine to assign, column left blank); and the outbound policy decision itself (plausibly intentional given self-hosted SearXNG — needs a product answer, tests specified either way).

**One correction back:** your "seven on Windows, eight on POSIX" — the Windows figure is measured, the POSIX one isn't, here or in your review. §8 step 1 makes it a task rather than an assumption.

Separately, `SERPER_API_KEY` is now configured as an Actions secret, so §6/§10.3 enable `live-serper` and `live-extraction` (the latter needs a browser but no credential) nightly. The four providers without credentials stay **out** of the matrix rather than skipping green — adding a secret is what adds the job. Fork-PR secret handling is stated explicitly and checked against GitHub's `pull_request_target` guidance.

shelpuk:
Revision 3 pushed (`3d00488`), and the PR description is rewritten — it was describing revision 1 throughout, including the "no new spend" line, which is wrong now that a billable nightly canary exists. Thanks for catching that; it was the summary everyone would have read.

I verified each factual claim against the tree before acting. Four changed the design:

- **`outputSchema` is `null` for both tools.** Confirmed — both are annotated `-> dict`, so FastMCP derives nothing. §4.2 now separates the three things you identified: the generated schema clients actually receive, the Pydantic model schemas, and runtime result validation. It pins `outputSchema is None` as a deliberate assertion, so it fails the day someone changes the annotations. Annotating them properly is listed as an API decision in §13, not smuggled in as a test change.
- **`pytest -m nonexistent_marker` exits 0**, not 5 — I checked, and it reports "7 deselected". Worse than your framing: I'd have relied on exit-5 as the backstop, and there isn't one. Every marker-selected job now needs an explicit minimum-collected-count step.
- **`fetch_html_via_nodriver` hardcodes `base_cmd` at `universal_html.py:550`.** Confirmed. There's now a §11 listing the production seams this design depends on, so they get scheduled rather than improvised as monkeypatches — which is exactly how `_FakeProc` drifted in the first place.
- **`test_serper_live.py` uses `urllib.request` directly.** Confirmed — it bypasses `PROVIDERS`, `search_web` and the tool entirely, so my claim that the canary covers the routing path was false for the test that exists. §6.2 now requires the public surface.

**On the sandbox helpers** — you're right that I mislabelled them pure. `_resolve_sandbox_enabled` reads `os.geteuid()` and env; `_resolve_browser_executable_path` reads four env vars then probes PATH via `shutil.which`. §3.1 specifies the seams instead. And your sharper point stands: those tests also assert retry and termination *orchestration*, which can't move to a resolver test. §8A splits them rather than relocating the file.

**Marker collision** — caught and fixed. `live-extraction` would naturally carry `chromium`, so the PR job now selects `chromium and not live`, and installed-wheel tests live in `tests/package/` as well as carrying a `package` marker, so a mis-set marker can't include them against the source checkout.

Also applied: CI axes for Python 3.13/3.14, mcp min/max and `pdf-advanced` present/absent with cost rationale; an operational ratchet definition (combination set, merge-base comparison, path normalization, tolerance, exclusions) plus your correction that a report reveals the gap while a global ratchet might not; one canonical valid initialize request so a protocol 400/406 can't masquerade as a security result; sanitization before `entries` is appended, not just at the writer — that was the sharpest of the low-severity items, since `entries` is the exposed path; async migration decoupled from baseline restoration; identifier-preservation properties replacing the undefined round-trip; corpus snapshots added on regressions and deliberate changes too; and the matrix "Today" column made consistently about coverage, with inbound transport marked as an automated gap.

**Structure** — agreed, and acted on. The revision history and rebuttals are gone; §13 is now "Open decisions" listing the four things that need your answer rather than arguments with a previous draft. The narrative lives in the PR description where you suggested.

**One I'm not taking:** the `.requirements/` point. This project's convention is a per-task `.requirements/<datetime>_<feature_name>/REQUIREMENTS.md` at the repo root, not a single root `REQUIREMENTS.md` — §14 refers to the directory because that's the convention that's missing. Happy to be corrected if you're reading from a different standard, but I'd rather the doc match the workflow this repo actually uses.

The most consequential thing still open is §13.1: whether fetching private-network addresses is intentional. That's a product answer, and the security section can't be finished without it.

shelpuk:
Revision 4 pushed (`db0c856`). All nine points applied — and testing your H1 turned up an error of my own that I should flag prominently.

**My "exits 0" measurement was wrong.** `pytest -m nonexistent_marker` exits **5**, not 0. I had piped pytest into `tail`, so `$?` reported `tail`'s status. Re-measured with the exit code captured directly:

| Command | Exit |
|---|---|
| `pytest -m typod` (nothing matches) | **5** — already fails CI |
| `pytest -m slow`, 1 selected where 5 expected | **0** — passes silently |

So the dangerous case is **under-selection, not zero-selection** — a renamed marker left on two files out of forty. The guard is still worth having; the justification I gave for it wasn't. The doc now states the measured behaviour of both cases.

**Your H1 stands regardless, and the fix is in.** `pytest … || [ $? -ne 5 ]` is broken exactly as you describe — it lets an under-selected run pass *and* converts a real exit-1 failure into a green job.

**The replacement had a bug too, which testing caught.** `pytest_collection_modifyitems` runs *before* mark-based deselection, so it sees the full collected list and never fires — I verified it silently passing on a zero-match selector. It's `pytest_collection_finish` reading `session.items`. Verified: under-selection exits 4 with the count in the message, a real test failure keeps exit 1.

**H2 — agreed, and it was the right catch.** The original failure was an unenforced red suite, so deferring the gate to the end would let that recur while B–D are in flight. Enforcement is now workstream **B**, immediately after green: one `fast` job plus branch protection, with later jobs joining `ci-required` as their tests appear.

**H3 — fully specified.** You're right that stubbing search isn't enough; a non-empty result set fans out to `resolve_page_content_markdown` per link and launches Chromium. Two deterministic calls now: `web_search` against a zero-result fixture (explicitly labelled a routing/protocol/packaging smoke test, no resolver), and `get_content` on a `.pdf` URL — `load_url_as_markdown` calls `_is_probably_pdf_url` as its first statement and returns `None` before any probe or browser, so it's deterministic with no network and exercises the `None` → Markdown-note conversion §4.4 pins.

**M4** — you're right that I scoped it wrongly; session init, transport, invocation and entrypoints are all SDK-dependent, and SDK drift is exactly what `test_dependency_constraints.py` was written for. `package` carries the {min, max} axis now.

**M5** — rather than patch the definition I replaced it. `diff-cover --fail-under=80` is the gate; a committed baseline that may not decrease is the anti-erosion control. That drops the merge-base double-measurement along with its unanswerable "does the base run its own tests or head's?", and drops the 0.10 tolerance, which as you note compounds. Lowering the baseline now means editing a tracked file in the PR, which is visible in review — the actual goal.

**M6** — right, path and marker are different failures. Source jobs pass `--ignore=tests/package`, the package job targets the directory, and a collection-policy test asserts every file there carries the marker.

**L8, L9** — both in. `ci-required` is the single required check so matrix changes don't drop names from branch protection. The worker seam is a private `_run_worker_command`, with `fetch_html_via_nodriver` always building its own command — your point about not turning arbitrary command execution into part of the fetch API is well taken, especially on a function whose URL input is already attacker-influenced.

**M7 — I checked, and this repo has no `AGENTS.md`** (nor a root `REQUIREMENTS.md`, nor `.requirements/`). You may be reading a workspace-level file outside the checkout. Rather than argue conventions in a test document, §14 now just states the observable fact: no per-task requirements artefacts of any kind exist, so the acceptance scenarios are reverse-engineered either way. Which convention *should* apply is worth settling, but not here.

Still open and needing your call, unchanged: §13.1, whether outbound fetching of private-network addresses is intentional.

shelpuk:
Revision 5 pushed (`afbe6fb`), PR description updated. All seven applied. The three high-severity findings had a common shape — a control that could report green while the thing it protects was failing or skipped — which is the same class of problem as the original red-suite normalisation, so they were worth catching.

**H1 — right, and it undercut the whole point of moving enforcement earlier.** Workstream A restores the retry, cleanup and streaming tests, and §5.2 puts those at the subsystem layer, which `fast` excludes by construction. `ci-required` would have been green with exactly the tests that prompted this document unprotected. The first gate now selects `not live and not chromium and not package` on both platforms — excluding only what the first runner genuinely can't do.

**H2 — confirmed against GitHub's documented behaviour** ("if a job fails, all jobs that need it are skipped unless the jobs use a conditional statement that causes the job to continue"). The aggregator now uses `if: ${{ always() }}` and asserts every dependency's result is `success`. I took your stricter form over the common `!contains(needs.*.result, 'failure')` idiom for the reason you gave: the loose version accepts `skipped`, which is precisely how a required job that quietly stopped running goes unnoticed. Nightlies are out of `ci-required` and get their own summary, and I fixed the two contradictory statements about which checks are required — §10.3 was right, the later paragraph was left over.

**H3 — you're right that what I wrote was a floor.** Both holes are real: an unrecorded rise can be given back later, and lowering the file in the same PR was only *visible*, not *blocked*. Now three checks — `head_baseline >= base_baseline` read from the PR base via `git show`, `measured == head_baseline` at a stated precision so rises must be recorded, and structured `coverage json` rather than parsed console text. That gives a genuine non-decreasing guarantee for the cost of one `git show`, which is what I was trying to buy with the stored number in the first place.

**M6 — verified, and worse than a theoretical concern.** `parse_arxiv_url("https://arxiv.org/pdf/2401.12345.pdf")` accepts and returns `2401.12345`, so my "deterministic, no network" wheel smoke test would have hit arxiv.org for any arXiv PDF. Pinned to `https://example.invalid/package-smoke.pdf` (RFC 2606, cannot resolve), with a companion L1 test asserting every specialized parser rejects that exact URL — so it can't silently start reaching the network if a parser's matching widens later.

**M4** — `import pytest` added. Your second point is the more valuable one and I'd missed it: a floor detects loss but not omission, so ten new unmarked tests slide under a minimum of forty. The doc now commits the minima with a rule to update them on any deliberate change, and says to prefer a marker-policy test wherever ownership is structural — `tests/package/` being exactly that case, and strictly stronger than a count there.

**M5** — specified: which job emits `coverage.xml`, `--compare-branch=origin/main`, `fetch-depth: 0` for the shallow-checkout case, and partial-branch handling. Also softened the claim — an 80% threshold promises "at least 80% of coverable changed lines", not "new code is covered or the PR is red". That was overselling it.

**L7** — updated, and I removed the line count from the description entirely rather than correcting it again. It's been the recurring staleness source across three revisions and carries no information a reader needs.

Still open and unchanged: §13.1, whether outbound fetching of private-network addresses is intentional. That's the one item blocking the security section, and it needs a product answer rather than another revision.

shelpuk:
Revision 6 pushed (`3ea461d`). All seven applied. Two were claims the tools can't actually deliver, which is the worst kind of error in a document meant to be implemented from — thanks for catching both.

**H1 — right, and the contradiction was self-inflicted.** I pinned `diff-cover>=9,<10` and then described branch-aware behaviour in the same section. Raised to `>=10.4,<11` so the option exists, but I've taken your option (b) for the description: the gate is changed-line execution only, with branch completeness left to the whole-project report and mutation testing. I can't install here to exercise `--branch-coverage`, so the doc explicitly does *not* assume it until someone does — overclaiming it once was enough.

**H2 — you're right, and the failure mode you describe is the important part.** A fixed job *selection* isn't a fixed *environment*: the moving "newest allowed" mcp, unpinned app deps, and a Chromium package that changes underneath the job all move the number on an unchanged commit. Under exact equality that means unrelated PRs go red, or authors start editing the baseline to make CI pass — and a baseline people routinely edit is no longer a control at all. The ratchet now runs in a pinned lane (pinned `requirements.txt`, exact mcp, container by digest, `==` on the coverage tooling, 3.13 only) and the compatibility axes are excluded from it. Useful side finding: the repo already has a fully pinned `requirements.txt` from `pip freeze`, so the lane has something to install from today. (Also noticed `uv.lock` is referenced in `pyproject.toml` and `test_dependency_constraints.py` but doesn't exist — unrelated, not touched.)

**M5 — verified, and you were right that it was an overclaim.** `create_autospec(asyncio.subprocess.Process, instance=True)` declares `returncode`, `wait`, `kill`, `terminate` — and **not** `stdout`, `stderr`, `stdin`, `pid`, since those are set on the instance. It would have accepted the exact `_FakeProc` that was missing `stdout`. So the process double is pinned by a `WorkerProcess` Protocol with a runtime conformance test, and the real-child test stays. I kept autospec where it genuinely works — on the callables being patched, where it does catch signature drift — and said so explicitly, so the two recommendations don't read as contradictory.

**M3 — fixed properly.** Comparison moved into the Actions `if:` over named `needs.<job>.result` values, with the step reduced to `exit 1`. You're right that `toJSON(needs)` carries outputs, not just results, and that substitution happens before the shell parses the line. It was also dumping every dependency's outputs into the log for no reason.

**M4** — bootstrap rule added (baseline absent on the base SHA → skip check 1, require only equality; explicitly *not* treated as zero, which would let any value through as an improvement). Reads `github.event.pull_request.base.sha` rather than `origin/main`, and defines pushes to `main` and PRs targeting other branches.

**M6** — L4 now carries §5.4's rules plus the two you named. The import-resolution assertion is the one I'd have most regretted missing: without it the job can pass while exercising `src/`, which is a false positive on the single thing that job exists to prove.

**L7** — `--ignore=tests/package` added to the workstream B command, for exactly the reason §10.3 gives: the marker is the thing that gets forgotten.

Six rounds in, the findings have moved from strategy to tool mechanics, and the last two rounds have been me asserting behaviour I hadn't exercised. I've now verified every mechanical claim I could reach from here; the ones I can't — `--branch-coverage`, the Actions expressions — are flagged in the text as needing exercise at implementation rather than stated as fact.

Still open and unchanged: **§13.1**, whether outbound fetching of private-network addresses is intentional. That needs a product answer, not another revision.

shelpuk:
Revision 7 pushed (`9e10293`), PR description updated. All eight applied.

**H1 — this was the real one, and both horns of it are correct.** I checked the Dockerfile: `FROM python:3.13-slim` (moving tag), `apt-get install chromium` from Debian's rolling archive, and `COPY src` before `pip install .`. Rebuilt per PR the browser floats, so the lane isn't pinned; reused by digest it holds a previous revision of the application and can go green **without executing the PR's code at all** — which is the worse failure, because it looks like a pass rather than an error.

Taken your first option: a separate **test-runtime image** with only Python, Chromium and system deps, pinned by digest, into which the PR's wheel is installed at run time, with §6.1's import-resolution assertion applied so stale baked-in code can't satisfy it. The shipped `Dockerfile` is untouched — worth saying explicitly, since the product image and the test image now have different jobs.

**M3 — verified, and you were right about both halves.** Measured:

| Check | Missing `stdout` | Wrong `wait()` signature |
|---|---|---|
| `isinstance` without `@runtime_checkable` | `TypeError` | `TypeError` |
| `isinstance` with `@runtime_checkable` | caught | **not caught** (returns `True`) |
| static type check | caught | caught |

So the Protocol alone would have caught the missing `stdout` but not `_fetch_html` gaining five arguments — the other half of the original outage. There's now a narrow `mypy` job wired into `ci-required`, plus an explicit runtime contract test for attributes, callability and coroutine-ness. I scoped `types` to the Protocol-carrying modules rather than the tree: repo-wide checking of ~6,849 largely unannotated lines is its own project, and this job needs to earn its place immediately.

**M2** — real deadlock, and I'd left it to whoever hit it first. Now explicit: the default is that the author recovers the difference; a reset is permitted for measurement-semantics changes behind a label plus CODEOWNERS review of both `coverage-baseline.json` and the lane lockfile. Your framing was the deciding argument — a baseline people can edit at will isn't a control.

**M4** — caught a genuine inconsistency I introduced in the same revision: baseline on the event base SHA, diff coverage on `origin/main`. A stacked PR would have been charged for its parent's lines. Both now use `git diff <base-sha>...HEAD` through `--diff-file`.

**M5** — right, and the hole was one level down from where I'd looked. `requirements.txt` has no pytest/coverage/diff-cover, and `==` pins on those still leave their transitive deps free. Replaced with a committed `requirements-ratchet.txt` covering runtime and tooling together, regeneration command recorded. Hashes noted as worth adding if the lane is ever treated as security-relevant.

**M6** — correct, "require a pull request" is a separate setting from required status checks and bypass actors exist. Either it joins the required configuration, or a `main` push ratchets against its first parent's baseline. Recommended the first, documented the second so the rule isn't silently void.

**L7** — fair; the justification was circular. Kept the 10.4 floor but justified on its own terms: the lane pins tooling exactly, so starting lower would force a coordinated pin-bump-plus-baseline-reset later purely to gain an option. Adopting `--branch-coverage` stays a separate change that must be exercised first.

**L8** — updated, and I removed the round count the same way I removed the line count, since both have gone stale every revision and neither tells a reader anything.

Seven rounds. The one thing that hasn't moved is **§13.1** — whether outbound fetching of private-network addresses is intentional. It's a product answer, and the security section stays incomplete without it.

shelpuk:
Revision 8 pushed (`16e8f29`). All six applied — but by **simplifying**, because three of them were symptoms of one thing: the ratchet was scoped too widely, and §10.4 had grown to dominate the document for the least valuable control in it.

**H1 is the find of this round, and it contradicted a claim made earlier in the same document.** §10.4 said coverage exists to spot modules nobody tests, "which is how `chromium_pool.py` should have been noticed". With no `source_pkgs` set, it would not have been: coverage.py reports only files it observed executing, so a module no test imports is *absent* rather than visibly at zero. The one gap I used to justify the control was the one it couldn't see.

Fixed with `source_pkgs`, `relative_files`, the `[paths]` mapping — your point about `site-packages` versus `src/` is the subtle half, since unmapped the wheel-installed lane and the git diff never join and `diff-cover` silently analyses nothing while passing — plus the two assertions you specified.

**M2 — I took the underlying point rather than the suggested procedure.** You're right that pinning removes environmental drift but not timing nondeterminism, and that real Chromium, subprocess, retry and concurrency tests can cover different branches run to run. Rather than measure and hope, the ratchet is now scoped to the **hermetic `fast` lane only**. `subsystem` and `chromium` coverage is still reported; it just doesn't gate. Ratcheting on a number that flickers teaches people to edit the baseline until CI passes, which destroys the control outright. The stability criterion you described is in the doc as a precondition for enabling exact equality, stated on line and branch *sets* rather than rounded percentages.

That scoping also dissolves a coupling I'd created last round: no browser image feeds the baseline any more, so the test-runtime digest can't move it. The image requirement moved to §10.3 next to the `chromium` job, which is where it belonged — it governs how that job runs, not what the baseline measures.

**M6** — right, the exception was policy text next to an unconditional check, so an authorized reset would still have failed mechanically. Now a real condition: decrease fails unless the PR carries `coverage-baseline-reset`, the workflow triggers on label events so applying it reruns rather than needing a manual re-run, and the label sits behind CODEOWNERS on both the baseline and the lane lockfile, with stale approvals dismissed and CODEOWNERS self-owned.

**M4** — self-contradiction I introduced, fixed: `requirements-ratchet.txt` is authoritative everywhere. Your `pip freeze` catch is right too — after installing the project it records a local path or URL no other machine can resolve, so the regeneration command uses a dedicated `ratchet` extra and `--exclude-editable`.

**M3** — narrowed the promise to "changed statement lines represented in the coverage report", since continuation lines of multi-line statements are in the diff but not the statement-based report. `--expand-coverage-report` noted as worth adopting after validation on this codebase, not assumed.

**M5** — correct and I'd left it implicit. Added the forcing assignment, plus the two conditions that stop it passing vacuously: the fake must be fully annotated, and the job rejects `Any` on that surface — an `AsyncMock` member types as `Any` and satisfies any Protocol at all.

Net effect: §10.4 is shorter than before despite gaining a control, because narrowing the ratchet's scope removed more machinery than the fixes added.

**§13.1 remains the only thing I can't resolve by revision** — whether outbound fetching of private-network addresses is intentional. Eight rounds in, that's the one open item, and it needs a product answer.

shelpuk:
Revision 9 pushed (`47389f3`). All eight applied. The three highs were load-bearing — as written, **none of the coverage controls would have run**.

**H1** — there was no coverage job at all. §10.4 referred to a "combined-coverage job" I never defined, and `fast` can't stand in for it since it's a compatibility matrix while the ratchet needs one pinned environment. So a literal implementation gave all-green required checks having executed none of the three controls. Added `coverage` with its install method (source checkout, not wheel — the `[paths]` mapping is for the artefacts it *consumes* from `package` and `chromium`), selection, inputs and outputs, wired into `ci-required`.

**H2** — correct, and I'd have shipped it. `types:` replaces the defaults rather than extending them, so the workflow wouldn't have run on PR open. Worst on a fork PR, which produces no upstream `push` run either — and PR #50 came from a fork, so that path is live here.

**H3** — genuinely undefined, and you're right that both readings fail. Two products now:

| Product | Built from | Feeds |
|---|---|---|
| `coverage-hermetic` | the pinned lane's own run | the ratchet |
| `coverage-combined` | that plus every test job's artefact | module presence, diff coverage |

Diff coverage must use combined — charging a `chromium_pool.py` change against a lane that structurally cannot execute it would push authors to write the wrong test at the wrong layer, which is precisely the pathology §2.1 exists to prevent. That also corrects my previous wording: subsystem and chromium coverage **does** gate, via control 2. What it doesn't do is feed the ratchet. The distinction holds because control 2 asks a per-line yes/no at an 80% threshold, which jitter rarely flips, while control 3 compares an aggregate where one differently-taken branch changes the answer.

**M5 is the one I'd have been most embarrassed to ship.** Running the parent under coverage doesn't instrument `python -m …nodriver_worker`, so the worker would have read as uncovered no matter how well §5.2 exercised it — and once diff coverage uses the combined product, a PR touching the worker gets charged for lines it cannot cover. Added `parallel = true` + `patch = subprocess`, bound raised to `>=7.10,<8`. I also documented the limit you flagged rather than papering over it: a forcibly killed child cannot flush, so the kill-path tests contribute nothing by construction — and the answer is explicitly *not* a coverage exclusion, which would also hide genuinely untested code.

**M6** — right, and my flags were the wrong ones. `disallow_any_explicit` covers written-out annotations, `disallow_untyped_defs` covers definitions; neither catches inferred `Any`. Now `disallow_any_expr`, a concrete fully annotated fake instead of `AsyncMock`, and your negative fixture — a type-check job that cannot fail is indistinguishable from no job.

**M4** — the guard now fires only when the diff contains at least one represented statement line, so a literal edited on a continuation line isn't misdiagnosed as broken wiring.

**M7** — restored, and stronger than before: integer totals from `coverage json` (never `percent_covered`, which is a binary float), the combined statement-and-branch measure named explicitly since "branch coverage" is ambiguous, comparison in integer basis points, and a schema keeping raw counts beside the ratio so a reviewer sees *what* moved.

**L8** — correct, and I'd overstated it. Authorization is on the merge via required code-owner review; the label only unblocks the check, and anyone with write access can apply one. Reworded, with actor-verification noted as extra machinery this design doesn't require.

Nine rounds, and this round's findings were all in §10.4 and its CI wiring again. That's now a consistent signal: the coverage apparatus is generating most of the residual risk in a document whose own ranking puts it below mutation testing and diff coverage. If the next pass lands there again, I'd suggest cutting control 3 entirely rather than continuing to harden it — controls 1 and 2 plus mutation testing cover the stated goals, and the ratchet is the piece that keeps needing scaffolding.

**§13.1 is unchanged and still needs a product answer.**

shelpuk:
Revision 10 pushed (`a1d702f`). All eight addressed — but **H1, H2 and M4 are resolved by deletion rather than specification**, because H3 settles them.

**H3 is the decisive finding and it generalises further than stated.** You're right that on a five-statement diff, one timing-dependent line moves 80% → 60%. That means a required threshold gate cannot take nondeterministic input — and "rarely flips" was hand-waving on my part in the previous revision. A flaky required check is exactly the normalised-red failure this whole document exists to prevent, so I can't build one on the same reasoning I rejected elsewhere.

Once that holds, the cross-job pipeline has no purpose left:

- Control 1 never needed it — `source_pkgs` enumerates never-executed modules from any single run.
- Control 2 can't soundly use it, per your H3.
- Control 3 already excluded those lanes.

So all three controls now measure one hermetic run, and the `coverage` job consumes no artefacts. That removes the whole failure surface you enumerated in H1 — unique names per matrix leg, hidden-file exclusion, download collisions, combine's replace-vs-append behaviour — along with H2's missing `fast-extras` leg and M4's cross-producer configuration alignment. Nothing can silently reduce the measurement to the hermetic lane, because that's all there is.

**The cost is real and I've stated it rather than buried it:** changes to code reachable only through `subsystem` or `chromium` are no longer diff-gated. What covers them is §2.1's allocation rule in review, those jobs being required, and control 1 still failing if such a module has no tests at all. Their full coverage is still produced and read as observational.

**M5** — took the sharper bound. `>=7.10.3,<8`, pinned `==7.13.5` in the lane. The 7.10.3 fixes you list — nested children, data stranded on a child `chdir`, Windows startup, incomplete config propagation — are not incidental here; the worker tests hit all four.

**M7** — you caught a straight contradiction: I leaned on mutation testing for killed-worker paths while §3.2 explicitly excludes `scrape/` from `mutmut`'s scope. Claim withdrawn; the real-child assertions carry it, and widening mutmut's scope is noted as a separate decision with its own cost.

**M8** — `basis_points`, `covered` and `total` must all match. Your point about compensating changes decided it: a file removed and another grown can hold the ratio while both raw numbers move. `generated_by` is validated separately so a tooling bump reports as a tooling mismatch rather than a coverage regression.

**M6** — all three fixed, and the trigger one is instructive: the duplicate had drifted into the incomplete form that would have stopped the workflow running on PR open. The reset section now defers to §10.3 rather than restating it, with a note saying why.

On process: this is the third consecutive round inside §10.4, and the second where the right answer was to remove machinery rather than add it. Last round I predicted the next findings would be about control 3 and suggested cutting it — that prediction was wrong. The complexity was in the *cross-job fan-in* feeding controls 1 and 2, not in the ratchet. Removing it took out more than cutting control 3 would have, and kept all three controls.

**§13.1 remains the only open item that a revision cannot settle.**

shelpuk:
This review targets `47389f3`, which is one commit behind — the head is now `a1d702f`, and all eight findings are addressed there. My previous comment covers the reasoning; the short mapping, verified against the current file:

| Finding | Status on `a1d702f` |
|---|---|
| 1. Artifact fan-in not executable | **Removed, not specified.** The `coverage` job consumes no artefacts: "No `needs` on the test jobs, no artefact upload or download, no cross-job `coverage combine`" |
| 2. `fast-extras` missing from combined | **Moot** — there is no combined product |
| 3. Nondeterministic lanes unsafe for diff coverage | **Accepted, and it drove 1/2/4.** Your 80%→60% arithmetic is quoted in the doc as the reason |
| 4. Producers need identical tooling/config | **Moot** — one producer |
| 5. Coverage bound includes buggy versions | `==7.13.5` pinned in the lane, `>=7.10.3,<8` elsewhere, with your four 7.10.3 fixes named |
| 6. Stale statements | "required dependency of `ci-required`"; summary lists `types` and `coverage`; reset section defers to §10.3's canonical trigger |
| 7. Mutation testing can't carry killed-worker paths | Claim withdrawn — "Mutation testing cannot help here" |
| 8. Baseline equality ambiguous | `basis_points`, `covered`, `total` all compared; `generated_by` validated separately |

Finding 3 was the one that mattered: it made findings 1, 2 and 4 disappear rather than need answering. A required threshold gate can't take nondeterministic input, so the cross-job pipeline that needed all that mechanical precision had no sound purpose left.

Happy to take a fresh pass against `a1d702f` whenever you're ready.

shelpuk:
Revision 11 pushed (`276182c`). The critical finding is correct, and it invalidated three separate claims — I'd asserted a scope I had no mechanism to produce.

`source_pkgs` makes the report whole-package **by definition**, which is the whole point of control 1. So `chromium_pool.py` isn't absent from `coverage.xml`; it's present with every statement at zero. Everything follows from that:

- `diff-cover` counts changed zero-hit statements as uncovered → "not counted for or against the gate" was false.
- L3-only statements grow the ratchet denominator while the numerator stands still → the baseline falls and blocks the PR.
- **Control 1 was tautological.** Under `source_pkgs` every module always appears, so "appears in the report" can only fail on a broken config. My claim that it "fails if such a module has no tests" was wrong, and worse, the control I'd built the section around had no real failure mode.

Ordinary worker or pool development would have been unmergeable. That's the kind of thing that gets a control switched off within a month.

**The fix makes the scope declared and checked** rather than wished for: one run, two report views — whole-package for control 1, `.coveragerc-gate` with an `omit` list for controls 2 and 3. Control 1 is rewritten to assert *coverage*, not presence: classification must be total and exclusive (policy test over the tree and the omit list), every gating-scope module must have non-zero coverage in the hermetic report, and every omitted module non-zero in the observational L3 report. That is what would actually have caught `chromium_pool.py`.

**Your `coverage-scope.toml` suggestion I took in substance but not in form.** A manifest is a side-table that can drift from the code. Since `omit` is file-granular anyway, the honest place for the boundary is a module boundary — so §11.2 now requires extracting `scrape/worker_runner.py` from `universal_html.py`, which today mixes a hermetically tested probe path with subprocess management. That was already half-proposed for the injection seam; the coverage scope makes it a prerequisite. Same argument §2.1 makes about assertions: the seam belongs in the design, not in metadata.

**H2** — right, and it was a straight documentation/reality mismatch: the displayed `.coveragerc` had neither `patch` nor `parallel` while the text said those jobs set both "in `.coveragerc`". Three configs shown now, each selected by absolute `COVERAGE_RCFILE` rather than discovery — §6.1 runs from outside the checkout where discovery fails — plus the local `coverage combine` you flagged, since `patch = subprocess` implies parallel files that reporting won't merge on its own.

**M3** — both contradictions fixed, and the `pip install --no-deps -e .` gap was real: `requirements-ratchet.txt` deliberately excludes the project, so nothing made it importable. Added, with an assertion that it resolves under the checkout's `src/` — the mirror of §6.1's assertion, catching the opposite error.

**M4** — the sharpest framing in the review, and I've quoted the reasoning back into the doc: unowned observational reporting decays into the signal nobody reads, which is this project's founding problem. Now artefacts with `if-no-files-found: error`, a per-module summary on the PR itself, and a named reviewer pending §13's owner decision — explicitly unowned until then rather than pretending otherwise.

**L5** — both fixed. The "deleting tests" one was a genuine inversion: deleting tests lowers coverage; it's deleting uncovered production code that inflates it.

**§13.1 is unchanged** and remains the one item no revision can settle.

shelpuk:
Plan revised and pushed (`bb4fa97`). All 13 findings applied. The four criticals were the same class of error — I'd collapsed "can implement", "can merge" and "can activate" into one arrow, and cycles hid in the ambiguity.

**The three cycles are now verified absent, not asserted.** `scripts/check_plan_dag.py` parses the per-step `Blocked by` declarations, rejects undefined references and fails on any cycle:

```
$ python scripts/check_plan_dag.py
OK: 61 steps, acyclic, longest chain 7
    startable immediately after E0-2: 23
```

That's your M11 — the prose graph is no longer the source of truth, and it can't drift again.

- **C1** — E2-1 is now test-only (Protocol, fake, negative fixture), E1-4 uses it for a green characterization baseline, E2-3 extracts against that, and a new E2-4 applies the annotation. Exactly the order you proposed.
- **C2** — split into E4-5a (build, not required) and E4-5b (activate, after E7-4). The deadlock you describe is real and nasty: `--min-selected` makes the job red, and the PR that fixes it can't merge past it.
- **C3** — E10-2 is now the assertion tool unit-tested against synthetic `coverage.json` fixtures, merging green; E10-3 points it at the real tree once the L3 tests land. You were right that "must fail on merge" contradicted my own rule that every merged step leaves the repo correct — that was a deliberate outage with a ticket number.
- **C4** — mutation configuration and its job ship together in E12-1; E4-7 creates the nightly with live jobs only.

**H5 is the one I'd most wanted pointed out.** The schedule contradicted the document's stated purpose — two of five engineers for a three-day "wave 1". Only E0-1 and E0-2 are genuinely serial, and the checker confirms **23 steps open the moment E0-2 lands**. Bootstrap is now hours followed by a nine-stream fan-out, with each stream starting on its own narrow dependency.

**H6** — a straight category error on my part. `_fetch_html` is the *child's* function (browser startup, connect retry, profile cleanup); `_run_worker_command` is *parent-side* (spawn, streams, heartbeat, termination). E1-3's tests stay where they are, and the plan now says so explicitly so nobody re-proposes the migration.

**H7** — E5-7 is scoped to the redaction helpers' current behaviour, which passes today (`test_diagnostics_masking.py`, 8 tests), so it merges green. Every emit-boundary assertion moved into E9-1 alongside the sanitizer that satisfies it.

**H8** — four independently mergeable batches plus a completion check, replacing one 18-file step that called itself atomic.

**H9** — new §3 registers the external prerequisites: outbound-policy decision, repository admin authority, container registry and credentials, live-query budget owner. Each has an owner, what it blocks, and a requirement to record a rollback procedure before activation. Only X-1 stops work starting; the others block activation steps, so the tests they gate can be written and merged first.

**M10, M12, M13** — E1-3 now follows E1-2 explicitly; the 500 ms startup budget becomes readiness polling with a generous timeout and duration as telemetry; corpus governance adds capture date and sanitation checks for cookies, tokens, PII and analytics payloads.

**§13.1 / X-1 is still the only thing that blocks work from starting.**

shelpuk:
Pushed `d934d0b`. All 11 findings applied.

**H4 first, because it undermines everything else I claimed last round.** You're right on every point, and I verified each:

- `impl E1-1…E1-5` → parsed as two of five steps.
- `E5-*` → parsed as no dependency.
- `X-*` → invisible, which is why "23 startable" wrongly included E9-2 and E4-5a.
- `longest_chain` sorted by blocker count, not topologically. Your nine-step chain was real; the declared graph actually contains a **ten**-step one it reported as 7.

A validator that guesses is worse than none because it gets believed — and I believed it. Rewritten to reject loose syntax rather than interpret it: `Blocked by` now has a grammar (`impl|merge|complete <ID>`, comma-separated), ranges and wildcards are errors, duplicate ids are errors, external prerequisites depended on with anything but `complete` are errors, and file ownership collisions are errors. `tests/test_plan_dag.py` covers all of those plus cycles, downstream-of-cycle, backward numbering and chain depth — 14 tests. Confirmed the old loose syntax is now rejected rather than silently misread.

```
OK: 63 steps, 4 external prerequisites, acyclic
    longest chain: 10 steps — E0-1 -> E0-2 -> E2-2 -> E2-3 -> E10-1 -> E10-2 -> E10-3 -> E10-5 -> E10-6 -> E10-7
    startable immediately after E0-2: 20
```

**C1** — the sharpest structural finding. The repo starts with 11 failures, so my "every merged step leaves the suite green" rule was violated by the first PR under it. Now two phases: before E1-6, no regression beyond the recorded baseline and nothing is required; after E1-6, fully green. E4-1 lands the workflow **reporting-only** — you're right that branch protection being off doesn't make a red workflow green.

**C2** — E4-1's broad job deliberately includes `subsystem` (TEST_SUITE §8B says the initial gate must), but I never added the step that narrows it to the normative `fast` selector. E4-2 now does that and adds `subsystem` properly; E4-3 adds `fast-extras`, which had no step at all.

**C3** — E10 now has the chain you specified: classification → checker on synthetic data → reporting-only coverage job → observational reports (which the omitted-module assertion needs, hence the E7-2/E4-7 dependencies I'd omitted) → diff coverage → ratchet → **E10-7 activates `coverage` in `ci-required`**, which nothing previously did.

**H8** — a literal implementation would have left `types` red forever. The negative fixture moves to `tests/typing_negative/`, excluded from the ordinary mypy target, exercised by a harness asserting non-zero exit *and* the specific diagnostic.

**H9** — E2-3 and E6-2 both touch `universal_html.py`'s diagnostics path. E6-2 now follows E2-3, and E7-2 depends on the final codec.

**H7** — all four fixed. E7-3 is written against a locally installed Chromium and only *merges* behind the registry image; E11 waits on the observable workflow rather than branch protection; E12-1 attaches to a new bare nightly foundation (E4-9) since mutation needs neither credentials nor canaries; and the stream map no longer claims coverage work starts on day one.

**H5, H6, M10, M11** — every row now has a Type (PR / milestone / decision / operation) and a deliverable with an acceptance check; non-PR steps are depended on with `complete`, so "merge E4-2" is gone. E11 batches enumerate their files and the validator enforces uniqueness. X-1 is the single outbound-policy decision with E9-2/E9-3 depending on it directly, and `outputSchema is None` is declared normative rather than "subject to" a pending choice.

Also took your closing note: the historical "an earlier draft wrongly…" narration is out of the plan. It belongs in commit messages, not in a document people are meant to execute.

Suite: **11 failed, 183 passed** — the same pre-existing 11, plus the 14 new validator tests, no regressions.

shelpuk:
Pushed `1f4bc84`. All 11 findings applied.

**C1 — you're right, and the fix is to delete the phase rather than shore it up.** A per-OS *count* can't distinguish fixing one stale test from breaking a different one, and "reporting only" doesn't make a red `ci-required` on `main` acceptable when the document says otherwise. Taking your second option: E4-1 is authored early but carries `complete E1-6` so it lands green, and E4-2 makes it required immediately after. That also fixes something I'd missed — my ordering delayed enforcement behind the job split and `fast-extras`, where TEST_SUITE §8B asks for the minimum gate right after restoration. E1-1 now records exact failing node ids too, which is the artefact that makes the remaining repairs checkable.

**H2 — the validator now enforces kind against target Type, and found all six mislabelled edges on its first run:**

```
FAIL: E4-2 declares 'merge E1-6', but E1-6 is a milestone and has no PR to land; use 'complete'
FAIL: E6-4 declares 'merge E11-5', but E11-5 is a milestone ...
FAIL: E11-1..E11-4 declares 'merge E1-6' ...
```

Model stated explicitly, as you proposed: `impl` = needed before authoring, any target; `merge` = PR targets only; `complete` = non-PR targets only. X-1 is now `impl` because it genuinely blocks authoring, and E7-2's `merge E6-2` became `impl E6-2` to match its own text.

**H3 — a real correctness problem, not just tidiness.** With the Protocol under `tests/`, E2-4 would have made `src/` import test code. It now lands in `src/.../scrape/types.py` as a type-only production change, and the fake imports it, so there is one definition rather than two that drift. E2-1's acceptance check asserts no module under `src/` imports from `tests/`.

**H4** — E9-2/E9-3 are now "enforce and test", conditional on X-1's outcome, sized `L` accordingly. You're right that there's no shared validation seam today, so a blocking policy makes a test-only PR unmergeable. Validation is specified at every redirect hop and at the resolved address, and E3-4 gains the injectable resolver seam so E9-3 is deterministic rather than racing DNS.

**H5** — my own rule, broken in two places. Each is split into add-job-reporting-only and a one-line activation.

**H6** — E0-1 now applies §10.2 in full, including bounds for the currently unbounded `pytest`, `ruff` and `packaging`, plus a `mutation` extra so the nightly has `mutmut` without putting a Linux-only tool in `dev`.

**H7** — E10-4 was doing six things. Split into worker observational, Chromium observational, the omitted-module assertion, and summary/delta reporting — the last now specifying artefact lookup, missing/expired-artefact behaviour, required permissions, fork behaviour (where `pull-requests: write` is unavailable, it writes to the job log rather than failing), and a new **X-5** for the reviewer and cadence §10.4 requires.

**M8, M9, M10** — E5-3 and E5-4 now own explicit disjoint function lists; E12-1's dependencies match §3.2's scope exactly (adding E5-4, dropping E5-6, whose accumulators aren't in scope) and it joins `nightly-summary.needs`; and the live jobs get separate credential criteria, since `live-extraction` needs no provider key.

**M11 — you were right that the document overstated the validator.** Rather than narrow the claim I extended it: kind-versus-type, rows that look like steps but fail to parse, empty file claims, claimed paths that don't exist, and `unittest`-style files claimed by no batch. That last check initially produced a false positive on its own test file, since the file contains the string `"import unittest"` inside a fixture — now matched as a real import at line start.

```
OK: 68 steps, 5 external prerequisites, acyclic
    longest chain: 11 steps
    startable immediately after E0-2: 20
```

21 validator tests. Suite: **11 failed, 190 passed** — unchanged baseline, no regressions.

shelpuk:
Pushed `21f82a4`. All 10 findings applied.

**H3 changed a headline number, so it's the one to check.** You're right that the metric contradicted §1.5: it excluded any step whose dependencies weren't E0-1/E0-2, regardless of kind, when the plan says only an unsatisfied `impl` stops authoring. Corrected:

```
OK: 72 steps, 5 external prerequisites, acyclic
    longest chain: 12 steps
    startable immediately after E0-2: 46      (was 20)
```

46 of 72, not 20. That is a material difference to how this distributes, and the plan had been understating its own parallelism by more than half.

**H1** — the ledger had no drain. Each repair step now deletes exactly the ids it fixes and adds none, and E1-6 checks the ledger *and* the suite together, since either can be true while the other isn't.

**H2** — two parts, and the second made me restate the rule rather than mechanically split more steps. `complete E4-2` is added to E4-3 so the ordering is in the graph. But on activation: the real invariant is that a job must not become *required* before its tests exist — not that every job needs a separate activation PR. Where the tests have merged and the runtime is ordinary, creation and activation in one PR is safe, because a job that can't pass simply doesn't merge. The split is now reserved for `chromium`, whose CI runtime is genuinely unproven. And `types` explicitly joins the aggregator, which it never said.

**H4** — you're right that "a single choke point" was wrong. E9 is now a shared policy function plus separate enforcement steps for `httpx` (every redirect hop), Chromium (its own stack), and deterministic rebinding — with X-1's artefact required to *select* between the allow and restrict subplans rather than just state a policy. The resolver seam moved out of E3-4 into **E3-6, labelled a production change**, since an effective injection point isn't a test helper.

**H5** — E10-6 implied exactly the cross-job fan-in TEST_SUITE §10.4 rejects. Each producing job now runs the checker against only its own report, with a per-job module list.

**H6** — correct, `$GITHUB_STEP_SUMMARY` needs only `actions: read`, which also deletes the fork special case I'd written. Noted explicitly that a PR comment would be a different design and isn't what the step builds.

**M10** — good call, and it removes X-3 from the authoring path entirely. The image splits into define-and-smoke-locally (no registry), publish, record digest; E7-3 merges against the local definition.

**M7** — grouped by behaviour class. "Parameterize over zero and negative" genuinely doesn't apply to `_cors_origin_regex` or `_detect_chrome_version`.

**M8** — a real false positive: `from unittest.mock import patch` is normal in pytest-native tests. Now matches framework use (`import unittest`, `unittest.TestCase`, `IsolatedAsyncioTestCase`) and not `unittest.mock`.

**M9** — added §1.3: each job lives in its own reusable workflow file, `ci.yml` only composes them, so the aggregator's `needs` list is the single contended region, and steps touching it are one line chained with `merge` where they'd collide.

21 validator tests; suite **11 failed, 190 passed** — unchanged baseline.

One thing worth naming: **X-1 now blocks five steps, not two**, and under a restrict policy three of them become production enforcement across three separate network stacks. It is the largest unpriced item in the plan.

shelpuk:
Pushed `cb5eca2`. All six applied.

**Your closing line was the finding**, and I've treated it as the organising one: every fix here moves a promise out of prose and into something checked.

**H1** — §1.3 claimed the aggregator edits were chained; the table didn't do it. The contended regions are now declared explicitly:

| Region | Mutated by, in order |
|---|---|
| `ci-required.needs` | E4-1, E4-3, E4-4, E4-5, E4-10, E4-11, E10-10 |
| `nightly-summary.needs` | E4-12, E4-13, E12-1 |

…the missing edges are in the graph, and `check_mutation_order` fails if any listed step doesn't depend on its predecessor. Your point about job definitions is taken: they live in separate files and stay parallel, so only the activations are chained — E4-9 can create the `chromium` job alongside anything.

**H2** — you're right that E9-5 could have gone green while neither path enforced. Split by transport: `httpx` rebinding behind E9-3, Chromium rebinding *and* redirect behind E9-4. The proxy point is the one I'd have missed entirely — `KINDLY_CHROME_PROXY` routes around host-side address validation by configuration, so it gets its own step.

**M3 — confirmed by reading it.** `test_live_fetch_urls.py` asserts only `"TimeoutError" not in page_content`, so it passes on an empty page or on the deterministic failure note. E8-5 now implements §6.2's thresholds, and E4-13 depends on it and on the runtime image, since that job needs a browser.

**M4** — resolved by ownership rather than a dependency: E8-4 must rewrite `test_serper_live.py` anyway, so it owns and migrates it and E11-1 drops it. Satisfying to report that the validator caught the resulting unclaimed file the moment I removed it from the batch — which is exactly what that check exists for.

**M5** — the floor is now a common acceptance requirement for every marker-selected job, with the maintenance rule for updating committed floors when a selection deliberately changes.

**M6** — the sharpest of the medium findings, because it meant the source of truth would go wrong the instant X-1 was decided. E9-2 now materialises the chosen subplan *in this document* (under an allow policy it deletes the four inapplicable rows and the validator must still pass), and a new **E13-1** milestone makes completion checkable — the validator reports what is listed, not what is finished.

```
OK: 76 steps, 5 external prerequisites, acyclic
    longest chain: 15 steps
    startable immediately after E0-2: 48
```

25 validator tests. Suite: **11 failed, 194 passed** — unchanged baseline.

One incidental fix: `FILES_LINE` required a line-start match, so E8-4's `Files:` declaration inside a bullet was silently ignored. Now accepts indentation.

shelpuk:
Pushed `f852bcc`. All six applied.

**H1 — and the check found more than the review did.** Rather than enumerate terminals by hand I added `check_terminal_reachability`, which fails if any step is not transitively required by E13-1. It reported **seventeen** missing, six beyond your list — E3-2, E3-3, E3-6, E9-2, E9-3, E9-4 as well. That difference is the argument for the check existing: hand-enumeration of terminals is exactly the thing that goes stale.

**H2** — you're right that a reusable workflow reports nothing until `ci.yml` calls it, so "reporting-only" still mutates `ci.yml`. `ci.yml.jobs` and `nightly.yml.jobs` are now tracked regions with their own ordering; the reusable-workflow *files* stay uncontended and parallel.

**M3 — and attempting the full fix surfaced a constraint worth recording.** `package` and the reporting-only `coverage` job now precede `chromium`, so the registry doesn't gate them, and `mutation` attaches to the nightly ahead of the live jobs. But chromium cannot be pushed back further: E10-5 measures the `chromium` job's own coverage and E10-6 asserts on it, so the coverage *controls* — though not the coverage job — genuinely depend on X-3. I found that by producing a cycle and having the validator reject it, which is a better outcome than reasoning about it.

**M4** — the figures were stale exactly as you say. They're now in a machine-readable comment checked by `check_declared_totals`, since a stale count is the one claim in the document a reader can't verify by reading. `test_real_plan_is_valid` would have kept passing otherwise, which was the real problem.

**M5** — correct: deleting E9-3 alone leaves E9-5's `impl E9-3` dangling and fails validation. E9-2 now specifies the whole transformation — four rows deleted, E9-5's dependencies and acceptance rewritten as characterization, E9-6 and E9-7 dropped from E13-1.

**M6** — right, `types` runs mypy and can't use a pytest option. Removed from that rule, with an equivalent non-vacuity check in E4-5: mypy's output must name every configured target module, so a job pointed at a path that no longer exists fails rather than passing with nothing to check.

```
OK: 76 steps, 5 external prerequisites, acyclic
    longest chain: 18 steps
    startable immediately after E0-2: 48
```

29 validator tests. Suite: **11 failed, 198 passed** — unchanged baseline. `git diff --check` clean.

The validator now checks eleven classes of defect. Every finding from the last three rounds is one it would catch today.

shelpuk:
Pushed `1aceebd`. All five applied.

**H1 is the most consequential finding in any of these rounds, because it changes what X-1 is.** I verified it in the source: the worker passes `--proxy-server=$KINDLY_CHROME_PROXY` straight through, and Chromium resolves and connects inside the browser process. So E3-6's Python resolver seam proves nothing there, and with a CONNECT proxy the *proxy* resolves the hostname — the application never sees the destination address, and a test navigating to a literal private URL doesn't exercise the bypass at all.

§3 now states that X-1 decides **two** things: the policy, and how enforcement is even possible on the Chromium path — with your four candidate architectures listed, including "trusted boundary, and document where the guarantee ends", which is the honest option if nobody wants to build a local policy proxy. E9-4, E9-6 and E9-7 now require the observable to be **the connection actually made**, not a pre-navigation classification, and E9-7 must use a hostname rather than a literal address.

**M2** — right, and the two rules interacted badly: E4-5 depended on E2-1, so its target list would have omitted `_run_worker_command`, and the non-vacuity rule I added last round would then have blocked adding a target that didn't yet exist. Now `merge E2-4`.

**M3** — the staffing number was inflated by four milestones and an operation. The validator reports PR authorability separately and the declaration carries `pr_authorable`; the plan names **44 PR steps**, not 48, as the figure to staff against.

**M4** — both stale figures removed rather than corrected. Numbers now appear once, in the machine-checked declaration, and the rest of the document refers to it. You're right that `check_declared_totals` only guards the declaration, so the fix is to stop repeating numbers rather than to check more places.

**M5** — `check_external_completion` now requires every declared prerequisite to gate a step E13-1 transitively requires, so a future external can't be added, block nothing, and be skipped while completion passes.

```
OK: 76 steps, 5 external prerequisites, acyclic
    longest chain: 18 steps
    authorable after E0-2: 48 (44 of them PR steps an engineer can pick up)
```

33 validator tests. Suite: **11 failed, 202 passed** — unchanged baseline. `git diff --check` clean.

Standing back: X-1 has grown across these rounds from "answer yes or no" to "choose a policy *and* a browser enforcement architecture". That is the right shape — the earlier framing was hiding work — but it means X-1 is not a five-minute decision, and it is the one thing gating a start date on the security stream. Everything else in the plan can begin.

shelpuk:
Pushed `2d4613a`. All five applied.

**H1's subresource half is the most serious security finding in this whole review sequence.** Everything in E9 assumed the navigation URL was the attack surface. It isn't: a permitted public page can reach `169.254.169.254` through an ``, `<script>`, `<iframe>`, a stylesheet `url()`, or `fetch()` — and blocking top-level navigation while allowing that is still SSRF. E9-4 now requires a fixture serving a permitted public page attempting each of those routes, with the private listener recording **zero** connections, and states plainly that a mechanism passing the navigation test while failing this one has not implemented the policy.

The other half is right too: "disallow the proxy" and "trust the proxy" answer what to do about an *upstream* proxy and say nothing about Chromium with no proxy configured, so neither can satisfy E9-6 alone. X-1 now decides three independent things, with the candidate mechanisms listed against the right question.

**M2** — the asymmetry was mine: I gave the allow branch a materialisation procedure and left the restrictive branch as one `M` PR that could mean a documented limitation or a whole local proxy with lifecycle, routing, resolver behaviour and auth chaining. X-1's artefact must now rewrite that branch into architecture-specific steps with their own sizes and acceptance criteria.

**M4** — and this one needs no timing at all, which makes it worse than the rebinding case I'd already covered. A name returning both a public and a private A record passes validation on one while the connector picks the other. E9-2 must now state whether every record must be permitted or the connection is pinned, with mixed record sets, dual-family answers and IPv4-mapped IPv6 in the matrix.

**M3** — both TEST_SUITE open decisions are now in the graph: **X-6** for the §9 risk-matrix owners (X-4 and X-5 named two; the rest were blank), and **E10-11** discharging §13.4's coverage-floor deferral once E10-3 produces the first real measurement. "Do not adopt floors" is a valid outcome with evidence; leaving it open indefinitely is not. Both are prerequisites of E13-1.

**M5** — checklist updated with the four checks added over the last two rounds.

```
OK: 77 steps, 6 external prerequisites, acyclic
    longest chain: 18 steps
    authorable after E0-2: 49 (44 of them PR steps an engineer can pick up)
```

The totals check caught the new step immediately on first run, which is exactly what it was added for.

33 validator tests. Suite: **11 failed, 202 passed** — unchanged baseline. `git diff --check` clean.

shelpuk:
Pushed `11c3efa`. All six applied.

**M5 was the finding that mattered, and H1 is its consequence.** You're right that the plan had drifted into asserting design of its own — its preamble says "not new design" while it had accumulated subresources, multi-address answers and a three-part Chromium decision that TEST_SUITE §7.2 did not contain. I've split that by policy-dependence:

- **The policy-independent attack surfaces move into TEST_SUITE §7.2 now**, because they are true whatever X-1 decides: multi-address answers, subresources, IPv4-mapped IPv6, and the `CONNECT`-proxy case where the proxy resolves and the server never sees the destination. §7.2 gains a fourth test boundary for browser-initiated requests and states that choosing the browser mechanism is part of §13.1's decision, not an implementation detail beneath it.
- **The architecture-specific parts wait for E9-0**, so the plan traces to the design rather than outrunning it.

**H1** — and the practical half is the one I'd have been caught by: every E9 step became authorable the instant X-1 was answered, so someone could pick up E9-4 before the re-planning existed. E9-0 is now a PR (`impl X-1`) that amends both documents, and E3-6 and E9-2…E9-7 depend on `impl E9-0` rather than on X-1 directly. Its acceptance requires the validator to pass afterwards, so a materialisation that leaves a dangling dependency fails.

**H2** — correct, and I'd listed the host-resolver rule as if it were a peer of the local proxy. It can't constrain a hostname the loaded page introduces, so it's now admissible only paired with per-request interception, and X-1's acceptance explicitly rejects any architecture mediating only top-level navigation.

**M3** — X-6 could have completed with names in a ticket and a blank Owner column in the design that is supposed to be authoritative. E13-2 writes them in, and E13-1 depends on that PR rather than the external.

**M4** — `impl E10-3`, not `merge`, since the decision can't be made before the measurement exists; and it becomes a PR because its artefact is an amendment to §13.4 with the measured figures.

**M6** — both fixed.

```
OK: 79 steps, 6 external prerequisites, acyclic
    longest chain: 18 steps
    authorable after E0-2: 49 (45 of them PR steps an engineer can pick up)
```

The totals check caught both new steps on the first run. 33 validator tests; suite **11 failed, 202 passed**, unchanged baseline; `git diff --check` clean.

shelpuk:
Pushed `abb27d1`. All five applied.

**Process change first, because it affects everything after this.** Deferred review points are now recorded **in the documents** as `> **Deferred:**` blocks naming what was raised and why it wasn't taken — not just answered here. Several items in these documents were rediscovered three and four times precisely because the reasoning lived in review threads that the next reader never saw. That convention is now in the plan's §1.7 and TEST_SUITE §13.

**H1 is correct and the failure is concrete.** The binary switch couldn't express what is probably the right answer: permit private HTTP(S) so self-hosted SearXNG and intranet docs keep working, refuse `file:`, `data:`, `ftp:` and `chrome:`. Under "allow" every enforcement step was deleted, leaving E9-2 classifying schemes with nowhere to enforce them. §13.1 now decides four things with schemes and address classes as **separate dimensions**, and enforcement is retained whenever *any* dimension restricts *anything* — only permitting every scheme and every address class removes the need for it. E9-0's materialisation is keyed the same way.

**M2** — my error from last round: I added E9-0 and left the materialisation block in E9-2, so after E9-0 lands E9-2's instructions are stale. Removed; E9-2 owns the policy function and its tests.

**M3** — "any security step" reads as including E9-1, the diagnostics sanitizer, which is independent of X-1. Both documents now say "outbound-policy step" and name E3-6 and E9-2…E9-7 explicitly. Worth catching: prose that over-blocks is as costly as a graph that does, because people follow the prose.

**M4** — the taxonomy was wrong in both directions: it claimed four decisions were unsettled when §13.2 is provisionally resolved and §13.4 is scheduled, and said the plan tracks "the same items" when X-2…X-5 are operational prerequisites appearing only there. Renamed to "Decision records and external prerequisites" with the four states tabulated.

**M5** — fixed; E9-4 is a placeholder, not a specification.

**First deferral recorded**, under §13.1: further policy dimensions — redirect depth, non-standard ports, method, response size — are deliberately not enumerated. Each speculative one adds a branch E9-0 must materialise and a column X-1's answer must fill, for a restriction nobody has asked for. The structure supports adding one in a paragraph if the decision needs it.

```
OK: 79 steps, 6 external prerequisites, acyclic
    longest chain: 18 steps
    authorable after E0-2: 49 (45 of them PR steps an engineer can pick up)
```

33 validator tests; suite **11 failed, 202 passed**, unchanged baseline; `git diff --check` clean.

shelpuk:
Pushed `6920c97`. All four applied.

**H1 is right and the framing is the fix.** I had enforcement and tests as one thing, so "nothing restricted" deleted both — leaving redirect handling, Chromium navigation and subresources, Chromium DNS and proxy behaviour with no characterization at all. E9-2 doesn't cover it: a pure function with no call sites says nothing about the server's behaviour.

They're separated now. Only production enforcement is conditional; where it isn't required the step stays as a **conformance test** asserting the route is permitted, or recording where a guarantee stops — the `CONNECT` proxy resolving on the server's behalf being exactly the limit worth pinning. A deleted boundary test isn't a lighter suite, it's an unmeasured one.

**M2** — and your example is the one that makes it concrete. Permitting private addresses while refusing `file:`, `data:`, `ftp:` and `chrome:` needs scheme enforcement on both stacks but **no** rebinding seam, no mixed-address handling and no proxy policy — a `CONNECT` proxy reaching a private address isn't a violation when private addresses are permitted. Collapsing that into "anything restricted" bought the maximal security implementation for a scheme restriction. E9 now carries a per-step table of which dimension's restriction requires its enforcement, and E9-0 applies that rule per dimension rather than choosing a branch.

**M3 is the one I should be most embarrassed by** — "three decisions" sat two paragraphs after a warning against restating the decision, and was wrong within one round of §13.1 growing a fourth. The count is as restateable as anything else, so it's gone entirely: the plan refers to §13.1 and summarises nothing.

**M4** — fixed.

**Second deferral recorded**, in the plan's E9 overview: enumerating every dimension combination. Two dimensions give four and each further one doubles them; the table gives E9-0 the rule instead, which doesn't go stale when §13.1 gains a dimension. That felt worth writing down rather than leaving as an absence someone re-raises.

```
OK: 79 steps, 6 external prerequisites, acyclic
    longest chain: 18 steps
    authorable after E0-2: 49 (45 of them PR steps an engineer can pick up)
```

33 validator tests; suite **11 failed, 202 passed**, unchanged baseline; `git diff --check` clean.

shelpuk:
Pushed `119402d`. All four applied.

**H1 was a contradiction I introduced last round**, and taking option (a) led somewhere better than just fixing it. E3-6 is now unconditional — the seam is small, defaults to today's behaviour, and exists for testability exactly like `_run_worker_command`. Its justification never depended on the policy.

Which surfaced a further point: **E3-6's API is identical whichever way X-1 goes** — only what attaches to it differs, tests scripting it under a permissive policy, enforcement wrapping it under a restrictive one. So it no longer waits on X-1 at all. That was false blocking on the one seam every retained rebinding test needs, and removing it opens one more PR step from day one:

```
authorable after E0-2: 50 (46 of them PR steps an engineer can pick up)
```

**M2** — correct, and it's the mirror of H1. E3-6 was conditional and shouldn't have been; E9-2 was unconditional and shouldn't have been. A permit-everything policy would have shipped a tested-but-unreferenced classifier, satisfying §1.7's letter while breaking its intent. Now dropped in that case, its cases folding into the boundary tests.

**M3 is the most valuable of the four**, because it guards the step everything else in E9 depends on. You're right that a rewrite could pass the DAG check, have no dangling references, update both documents — and still have lost a boundary or kept enforcement nobody asked for. All four of your semantic checks are now E9-0's acceptance, ahead of the syntactic ones, with a note that the syntactic checks are necessary and not sufficient.

**M4** — proxy is the fifth boundary now, with the reason stated: the other four assume the server can see the destination address, and the `CONNECT` case is precisely where it cannot. Treating it as a facet of the others was how it ended up owned by E9-7 alone.

Nothing deferred this round.

```
OK: 79 steps, 6 external prerequisites, acyclic
    longest chain: 18 steps
    authorable after E0-2: 50 (46 of them PR steps an engineer can pick up)
```

33 validator tests; suite **11 failed, 202 passed**, unchanged baseline; `git diff --check` clean.

PR #57, opened 2026-08-31
Record the per-OS baseline as a guarded ledger (plan step E1-1)

shelpuk:

This project's test suite has twelve tests that have been failing for a long
time. The team has learned to read past the red, which is a slow-acting problem:
once a suite is permanently red, a *new* breakage looks exactly like the old
noise, and nothing stops it merging.

Four upcoming tickets each repair a slice of that. The difficulty is that "twelve
failures went to nine" does not prove three were fixed — it is equally consistent
with four fixed and one newly broken. A count cannot tell those apart.

This PR builds the accounting record that can: a ledger naming each failing test
individually, per operating system, plus an automatic check that compares the
ledger against a real run every time the suite executes. From here on, each
repair must delete exactly the tests it fixed and no others, and the check
enforces it. When the ledger empties, the suite is green — those are the same
statement, so the milestone verifies itself rather than needing a separate
inspection.

One deliberate cost: the suite now takes about 26 seconds instead of 12, because
the check runs the suite a second time in a child process. That trade-off is
recorded in the ledger rather than left for someone to discover.

---

## What plan step E1-1 asked for

Measure the baseline on Linux — the design document predicted it from source but
had never run it — and record both platforms as **exact failing test ids**, not
counts.

## What was measured

Both platforms, at commit `3af0563`:

| Platform | Result |
|---|---|
| Linux (Ubuntu 24.04, CPython 3.13.15) | **12 failed**, 303 passed, 2 skipped |
| Windows (Server 2025, CPython 3.13.15) | **11 failed**, 303 passed, 3 skipped |

The prediction holds. Windows is the Linux twelve minus the root-detection test
it skips, and the two runs reconcile arithmetically: the same 303 passing, with
that one test counting as a failure on Linux and as the third skip on Windows.

The Windows number came from a throwaway `windows-latest` workflow on a branch of
its own, deleted as soon as the numbers were recorded. It touched neither
`ci.yml` nor any workflow region §1.3 serialises. **It is not the Windows CI
lane** — that belongs to the CI epic. It was an instrument, used once.

## What is in the diff

| File | Role |
|---|---|
| `.system_design/BASELINE_FAILURES.md` | The ledger — ids, counts, environment, commit, and the reasoning behind each choice |
| `tests/test_baseline_failure_ledger.py` | 15 checks; the load-bearing one re-runs the suite in a child process and compares |
| `tests/_baseline_probe.py` | Child-side plugin recording the run's outcomes |
| `.system_design/TEST_SUITE.md` | §1.1's POSIX figure changed from predicted to measured; §8 points at the ledger |
| `.github/review/rules/python-tests.md` | New guard registered as load-bearing; count corrected to eight |
| `pyproject.toml` | The `slow` marker's comment no longer claims it has no real user |

## Two findings from review that changed the design

**The first version could go green on a red suite.** It read results from
pytest's short summary. On pytest 9.1.1 a failing `unittest` subtest prints a
*passing* dot in the progress line and a `SUBFAILED` line in the summary — never
the word `FAILED`. A summary parser therefore reports an empty failing set while
the suite is red, in exactly the drained end-state this guard exists to certify.
Two of this repository's three `subTest` sites sit in a file the ledger tracks.
Results now come from a plugin reading pytest's own report objects, and the
child's exit code is cross-checked against what the plugin recorded, so the same
class of blindness fails loudly instead of passing quietly.

**An id could leave the ledger without being repaired.** Deleting a test,
renaming its class, or marking it skipped all remove it from the failing set and
all look identical to a fix. The child now reports its collected and skipped sets
too, and those three outcomes are named separately from a genuine repair.

## How the checks were shown to work

Every check was watched failing before being accepted, by mutating either the
document or the guard: deleting an id, adding one that does not exist, corrupting
the count, unsorting a block, dropping a platform section, half-draining a repair
across the two platforms, drifting the documented command, dropping a name from
the environment sweep, removing the probe plugin, skipping a documented test, and
polluting the parent environment.

That last one is not hypothetical: with `RUN_LIVE_TESTS=1` exported, an
un-normalised child run picks up a thirteenth failure from a live network test.
The child's environment is swept of everything this project reads, and a check
recovers those variable names from the source tree so the sweep cannot quietly
rot as variables are added.

## Verification

- Failing set **unchanged** from the pre-change baseline — the sorted `FAILED`
  lines diff clean. Counts move only by the 15 passing tests this PR adds:
  `12 failed, 318 passed, 2 skipped`.
- The review system's own 482 tests pass.
- `ruff check` clean on both new modules; `ruff format` clean.
- `scripts/check_plan_dag.py` exits 0.

### DIFF
--- a/.github/review/rules/python-tests.md
+++ b/.github/review/rules/python-tests.md
@@ -44,7 +44,7 @@
 
 ## The guard tests — treat these as load-bearing
 
-Seven tests exist to hold an invariant that nothing else enforces. A pull request
+Eight tests exist to hold an invariant that nothing else enforces. A pull request
 that changes what they guard, without changing them, is a finding; a pull request
 that *weakens* one to make a change pass is a critical finding.
 
@@ -57,14 +57,16 @@ that *weakens* one to make a change pass is a critical finding.
 | `test_pytest_configuration.py` | that `[tool.pytest.ini_options]` matches TEST_SUITE.md §10.5, that the running pytest actually loaded it, and that `strict_markers` / `asyncio_mode` have their effect and not merely their declaration |
 | `test_min_selected_guard.py` | that the `--min-selected` collection floor in `tests/conftest.py` matches TEST_SUITE.md §10.3 and still fires — an under-selected marker job exits 4 and names both counts, while a genuine failure keeps exit 1 and a collection error keeps its own diagnosis |
 | `test_coverage_configuration.py` | that `.coveragerc`, `.coveragerc-gate` and `.coveragerc-subprocess` match TEST_SUITE.md §10.4 and still behave — the base config reports an unexecuted module at zero, the gate config omits the modules with no hermetic seam, and the subprocess config captures a child process |
+| `test_baseline_failure_ledger.py` | that `.system_design/BASELINE_FAILURES.md` names exactly the tests that fail today — it re-runs the suite in a child process with the live-test opt-ins cleared and compares node ids, so a repair that forgets to delete its ledger entry and a new red test both turn it red, and for opposite reasons it names separately |
 
-Three of these are coupled: `test_min_selected_guard.py` and
-`test_coverage_configuration.py` both import `_section_body` from
-`test_pytest_configuration.py` rather than keeping their own copy of the
-fence-aware section bound, which exists because the naive heading regex was
-measured wrong — on §10.4's own blocks, whose opening comment sits at column 0
-and reads as a heading. A change to that helper's name or behaviour breaks all
-three guards, so it is not the private detail its underscore suggests.
+Four of these are coupled: `test_min_selected_guard.py`,
+`test_coverage_configuration.py` and `test_baseline_failure_ledger.py` all
+import `_section_body` from `test_pytest_configuration.py` rather than keeping
+their own copy of the fence-aware section bound, which exists because the naive
+heading regex was measured wrong — on §10.4's own blocks, whose opening comment
+sits at column 0 and reads as a heading. A change to that helper's name or
+behaviour breaks all four guards, so it is not the private detail its underscore
+suggests.
 
 Plus `test_worker_launch_args_redaction.py` for the subprocess command line — the
 same class of guard, one layer down.
--- a/.system_design/BASELINE_FAILURES.md
+++ b/.system_design/BASELINE_FAILURES.md
@@ -1,0 +1,196 @@
+# Baseline failures — the ledger
+
+**Status:** As Is. This file records the suite's red baseline exactly as it is
+today, one node id at a time, so that the repairs in E1-2 through E1-5 can be
+checked rather than believed.
+
+`TEST_SUITE.md` §1.1 quotes the baseline as counts. A count is not a checkpoint:
+`12 failed` before a repair and `11 failed` after it is equally consistent with
+"one test was fixed" and with "two were fixed and a third was broken". Only the
+set of failing node ids distinguishes those, so the set is what this file holds
+and what `tests/test_baseline_failure_ledger.py` asserts against a live run.
+
+## How this file drains
+
+Every entry below is a test that fails today and is expected to. **E1-2 through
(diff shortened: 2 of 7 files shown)


shelpuk:
## Reviewer findings — both answered in `3cf1151`

**1. Stale `slow`-marker docstring (`tests/test_pytest_configuration.py:489-490`) — accepted, fixed.**

Correctly identified as drift *this* PR created rather than a pre-existing gap. The docstring said the control case was the only user of the `slow` marker "until the suite grows a test that earns it"; this PR grows exactly that test, and I updated `pyproject.toml`'s comment for the fact without updating the docstring that comment was contradicting. Now names both users.

**2. `ENVIRONMENT_NAME_SHAPE` is uppercase-only — real, but the proposed fix is worse than the gap.**

The finding is correct on its own terms: a future lowercase-*only* environment read, with no uppercase counterpart anywhere in the tree, would escape the scan.

I measured the proposed remedy before rejecting it. Widening the shape to lowercase admits **139 more string literals, of which 137 have no uppercase counterpart** and are ordinary dictionary keys and field names. That converts a precise check into a 137-entry allow-list that nobody will maintain — rot arriving faster than the rot it prevents. The scan's value comes from a residue small enough that every entry can be justified; at 137 entries it has none.

So the shape stays, with that measurement recorded in the code as the reason rather than left as an unexplained choice. The case that actually exists is closed from the other side:

- `_child_environment()` matches **upper-cased**, so one listed name covers both spellings. `NO_PROXY` in the sweep already drops `no_proxy`.
- The check plants both `name` and `name.lower()` for every name found.
- `no_proxy` — the single lowercase variable this project reads — now has an **explicit control** of its own rather than incidental coverage.

Falsified rather than asserted: reverting the sweep to case-sensitive makes that control fail with a message naming the variable and the reason.

```
AssertionError: `no_proxy` survives the sweep. It is read at
scrape/nodriver_worker.py and scrape/universal_html.py alongside `NO_PROXY`,
and the sweep is case-insensitive precisely so that listing the upper-case
name covers both spellings.
```

The residual is stated honestly in the code: the scan covers uppercase-shaped names, and lowercase spellings are covered through the sweep's case-insensitive match rather than through the scan.

---

**Verification after both fixes:** failing set unchanged from the pre-change baseline (sorted `FAILED` lines diff clean) — `12 failed, 318 passed, 2 skipped`. Review system's 482 tests pass. `ruff check` and `ruff format` clean.

shelpuk:
## Accepted and fixed in `b5502e7`

Correct finding, and the fault was in how I *falsified* it rather than only in the code.

The control assertion sat after the `finally` that restores `os.environ`, so it read the ambient environment rather than a planted one. On any machine that does not export `no_proxy` — most of them, and every CI runner — it passed whether the sweep was case-insensitive or not. My earlier "falsified rather than asserted" claim only held because the command I falsified with happened to export `no_proxy=example.com`. A control that passes vacuously off the author's machine is precisely the defect this check exists to catch, one level up. The prose overstated it and the commit message did too.

`no_proxy` is now planted explicitly and the result captured inside the `try` block. Re-falsified the honest way — sweep reverted to case-sensitive, with both `no_proxy` and `NO_PROXY` **unset**:

```
$ env -u no_proxy -u NO_PROXY pytest ...::test_child_environment_drops_every_variable_the_project_reads
E  AssertionError: `no_proxy` survives the sweep. It is read at
   scrape/nodriver_worker.py and scrape/universal_html.py alongside `NO_PROXY`,
   and the sweep is case-insensitive precisely so that listing the upper-case
   name covers both spellings.
```

Your reading of the blast radius matches mine: the main planted-names check was never blind here — a planted lowercase `no_proxy` uppercases into the scanned set and would have surfaced as a survivor on any machine. What was broken is the specific named message the control exists to produce.

**Verification:** failing set unchanged from the pre-change baseline — `12 failed, 318 passed, 2 skipped`. Review system's 482 tests pass. `ruff check` and `ruff format` clean.

PR #59, opened 2026-09-01
Move the sandbox flag and default assertions to component level (plan step E1-2)

shelpuk:

The server renders JavaScript-heavy pages by launching a headless Chromium. Four
small decisions govern that launch: whether Chromium's security sandbox is on,
which browser binary to use, how many times to retry a failed start, and what
command line to hand it. Those decisions matter — one of them is the difference
between running a browser sandboxed and running it wide open on untrusted pages.

None of them was actually being checked. The tests that were supposed to check
them did it by running the whole 480-line browser-startup routine, and when that
routine gained five new required arguments, all eight tests began failing before
they reached a single assertion. They had been dead for some time and the suite
was red enough that nobody noticed which reds meant what.

This change asserts five of those eight claims against the small functions that
actually make the decisions. They now run in a quarter of a second with no
browser, no network and no subprocess — so they can run on every commit, and a
future change that flips a default or drops `--no-sandbox` turns something red
immediately instead of silently.

The remaining three tests assert retry and cleanup behaviour that genuinely
belongs to the big routine. They are untouched here and are the next step's job.

Suite: **13 failing → 7 failing**, 338 passing → 408 passing.

---

## What this implements

Plan step **E1-2**, "Move sandbox flag/default assertions to L1".

`tests/test_nodriver_worker_launch_resolvers.py` — 69 cases, 0.24s — asserts
`_resolve_sandbox_enabled`, `_resolve_browser_executable_path`,
`_resolve_start_retry_attempts` and `_build_chromium_launch_args` directly. The
five stale `_fetch_html` callers that used to carry those claims are deleted and
their ids drained from `BASELINE_FAILURES.md`.

Three of the five could not have been repaired in place even with the signature
fixed: they read `sandbox` and `browser_args` out of the keyword arguments given
to `nodriver.start()`, and the worker stopped passing either when it moved to
launching Chromium itself and connecting over the DevTools port. They were
asserting an interface that no longer exists.

### Hermeticity is the load-bearing part

None of the four functions is pure — between them they read eight environment
variables, `os.geteuid` and `shutil.which`. An autouse fixture pins all ten. That
is not tidiness:

- Without the `geteuid` pin, every sandbox case reports the right answer for the
  wrong reason inside a container running as root, and the root case proves
  nothing. Deleting the pin produces 6 failures under a simulated root.
- Without the environment sweep, `CHROME_BIN` exported by a CI image steers the
  browser-path cases. Deleting it produces 14 failures under a hostile
  environment.
- Without the `shutil.which` pin, the browser-path cases depend on whether the
  developer has Chromium installed.

### Validated by mutation, not by coverage

34 targeted mutations of the four functions' branches; **31 caught**. The three
survivors are each triaged in the docstring of the case that would have to kill
them, so the next reader of a mutant report stops there rather than writing a
test that cannot exist:

| Survivor | Why no test can kill it |
|---|---|
| `hasattr(os, "geteuid")` guard removed | The bare call raises `AttributeError` where the attribute is absent, and the surrounding `except Exception` catches it identically. Behaviourally equivalent for every input. |
| Negative spelling tuple narrowed | The resolver's fallthrough returns `False` too, so no input distinguishes "recognised as off" from "unrecognised, defaulted to off". **That branch is dead code today** — flagged, not deleted, since it stops being dead the day the default flips. |
| `.strip()` removed from the retry resolver | `int` tolerates padding, and an all-whitespace value raises `ValueError` into the same fallback the stripped empty string reaches. |

### Evidence that exactly five claims moved, and none was dropped

The ledger's rule is that an id may leave it only by *passing*, and the guard
enforces that by complaining about ids still listed whose tests have vanished.
Delete the id in the same change and the guard goes quiet — so the relocation
would otherwise be taken on trust.

The tests are deleted in one commit and the ledger drained in the next, so the
guard can be run in between. In that state it says, unprompted:

```
The live run disagrees with BASELINE_FAILURES.md on linux.
  In the ledger but no longer collected (deleted or renamed, not repaired;
  an id may leave this ledger only by passing):
    …::test_allows_enabling_sandbox_via_env
    …::test_disables_sandbox_by_default
    …::test_errors_when_no_browser_found
    …::test_forces_sandbox_off_when_running_as_root
    …::test_resolves_browser_executable_from_path
```

Exactly five, and no complaint of the other three kinds. The ledger now also
carries a table mapping each deleted id to the function its claim moved to, and
asks whichever of the next two repair steps lands first to make that table
machine-readable — they rewrite their tests, so their ids change too and they hit
the same wall.

---

## Changes outside the step, each with a reason

**1. `YDC_API_KEY` swept from the baseline guard's child environment.** `main`
was red beyond its recorded baseline: 13 failures, not 12. The You.com provider
merged after the ledger landed and reads `YDC_API_KEY`; the guard's honesty check
noticed the new variable and turned red, correctly. It is invisible to the
ledger's own live comparison because the child run excludes the guard's module,
so nothing else would have caught it. One line, first commit, so it can be
reverted independently.

**2. The ledger's `Result:` line is split in two.** Draining ids forces the
failure count to change — which turned a measurement into an arithmetic edit
sitting between three figures that were still the original run's. It read
`7 failed, 303 passed, 2 skipped` under a heading naming a commit where no such
run ever happened, and §1.1's rule is that baselines come from real runs or are
not quoted. `Result:` is now frozen permanently; a new `Remaining:` line carries
the count the guard asserts. The guard's synthetic documents state a deliberately
*different* number on the frozen line, so a parser reading the wrong field is
caught there rather than in the real document.

A visible consequence, left deliberately: the Windows section still says
`3 skipped`, and the third skip *was* the root-detection test this change
deletes. It is stale, it is labelled stale, and it is not corrected — editing a
measurement to keep it plausible is the exact failure mode the split prevents.
There is no Windows lane in CI, so it cannot be re-measured here.

**3. A claim was about to be dropped, and now has an owner.**
`test_errors_when_no_browser_found` asserted two things: that the resolver finds
nothing, and that `_fetch_html` turns that `None` into a `RuntimeError` naming
`KINDLY_BROWSER_EXECUTABLE_PATH` — the only actionable guidance a user gets when
no browser is installed. Only the first is a resolver claim. The second belongs
to `_fetch_html`, which this step is forbidden from calling, and no step owned
it: E1-3's four verify clauses are retry, terminate, non-retryable and cleanup,
and E5-4 takes resolvers only. It is now written into E1-3's scope and into §5.2.

**4. Two functions the design named as targets and no step claimed.**
`_is_snap_browser` and `_is_retryable_browser_connect_error` appear in §3.1's L1
target list but in no step's ownership. Given to E5-4 — a one-line plan edit that
widens no code here. Nothing checks §3.1's list against step ownership, so the
gap would not have surfaced a second time.

**5. E1-6 warned off a false green.** Draining the root-detection case made the
two platform blocks identical, which turns the cross-platform difference check
into a tautology. Combined with the Windows section being unmeasurable until
E4-1's matrix, the Windows block is now asserted by nothing that runs. E1-6's
verify clause now says its Windows half must come from a real Windows run, not
from the document.

**6. Documentation kept honest.** §1.1 is marked As-Is-frozen at `3af0563` (it
describes eight stale callers and twelve failures, which is history, not current
state) and its line-number citation to a now-deleted test is struck. §3.1 records
that the `hasattr` guard's mutant is equivalent, so a future mutation run's
report on it is triaged in advance, and names the new module next to the existing
precedent. Both reviewer rule files that carry this file list name the new module
— `scrape-browser.md` had the same list and said "four" — and record the split,
so a flag assertion is not added back to the file it just left.

---

## Verification

| Check | Result |
|---|---|
| Full suite | 7 failed, 408 passed, 2 skipped (from 13 failed, 338 passed) |
| Failing set vs. the ledger's Linux block | identical, byte for byte |
| Failing-set diff vs. `main` | six removals, **zero additions** |
| New module | 69 passed, 0.24s |
| Mutation battery | 31 of 34 caught; 3 triaged as unkillable, each in a docstring |
| Ledger guard, incl. live child run | 15 passed |
| Review system's own suite | 482 passed |
| `scripts/check_plan_dag.py` | exit 0 |
| `ruff check` on changed files | clean (`test_nodriver_worker_sandbox.py` improves 2 errors → 1; the remaining SIM117 is pre-existing) |

Reviewed by `system-design-reviewer` and `code-reviewer`; every finding from both
is either implemented above or recorded in the design with its owner.


shelpuk:
## Review addressed, and the evidence the review asked for

**The one suggestion is fixed** in `716cd42`. `TEST_SUITE_IMPLEMENTATION_PLAN.md`
said `test_worker_launch_args_redaction.py` "keeps only its redaction
assertions", which was imprecise: one of them asserts
`--proxy-server=<real credentials>` is present, pinning the flag's *emission* as
well as its content. That overlap is the point of the assertion — a redacted
copy means nothing without a real one to contrast — so it stays. What this step
owns is the branch *conditions*: when each proxy flag appears and when it does
not. Wording now says that.

**On "claimed, not shown."** Fair, and the fix is to show it. The subagent
reviews are not artifacts on disk, so here is the reproducible part — the full
mutation battery, re-run at HEAD. Each mutation is applied to
`src/kindly_web_search_mcp_server/scrape/nodriver_worker.py` in isolation, the
new module is run, and the source is restored; the run asserts the source is
byte-identical at the end.

<details><summary>34 mutations, 31 caught — full output</summary>

```
CAUGHT    sandbox: default flips to enabled  <- test_sandbox_falls_back_to_off_for_an_unrecognised_value, test_sandbox_is_disabled_when_the_variable_is_unset
CAUGHT    sandbox: root override removed  <- test_sandbox_is_forced_off_when_running_as_root
SURVIVED  sandbox: hasattr guard removed
CAUGHT    sandbox: identity catch narrowed to OSError  <- test_sandbox_request_is_honoured_when_geteuid_raises
CAUGHT    sandbox: affirmative spellings narrowed to '1'  <- test_sandbox_is_enabled_by_an_affirmative_value
SURVIVED  sandbox: negative spellings narrowed to '0'
CAUGHT    sandbox: value no longer lowercased  <- test_sandbox_is_enabled_by_an_affirmative_value
CAUGHT    browser path: explicit argument ignored  <- test_explicit_path_is_stripped, test_explicit_path_wins_over_environment_and_path
CAUGHT    browser path: explicit argument not stripped  <- test_explicit_path_is_stripped
CAUGHT    browser path: CHROME_BIN dropped from the tuple  <- test_a_blank_browser_path_variable_is_skipped, test_a_configured_variable_wins_over_a_browser_on_path, test_each_browser_path_variable_is_read
CAUGHT    browser path: precedence order reversed  <- test_browser_path_variables_are_read_in_declared_precedence
CAUGHT    browser path: env value not stripped, blank not skipped  <- test_a_blank_browser_path_variable_is_skipped, test_browser_path_variable_value_is_stripped
CAUGHT    browser path: probe order reversed  <- test_probed_browser_names_are_tried_in_declared_order
CAUGHT    browser path: probe list narrowed to chromium  <- test_each_probed_browser_name_is_tried
CAUGHT    browser path: PATH probe consulted before the environment  <- test_a_configured_variable_wins_over_a_browser_on_path
CAUGHT    browser path: not-found invents a default  <- test_no_browser_anywhere_resolves_to_none
CAUGHT    retry attempts: default 3 -> 4  <- test_retry_attempts_default_to_three, test_retry_attempts_fall_back_to_the_default_for_unusable_input
CAUGHT    retry attempts: floor removed  <- test_retry_attempts_are_clamped_up_to_one
CAUGHT    retry attempts: ceiling removed  <- test_retry_attempts_are_clamped_up_to_one
SURVIVED  retry attempts: .strip() removed
CAUGHT    launch args: --no-sandbox never emitted  <- test_no_sandbox_is_passed_when_the_sandbox_is_disabled
CAUGHT    launch args: --no-sandbox condition inverted  <- test_no_sandbox_is_absent_when_the_sandbox_is_enabled, test_no_sandbox_is_passed_when_the_sandbox_is_disabled
CAUGHT    launch args: profile directory dropped  <- test_the_profile_directory_is_passed_through
CAUGHT    launch args: user agent dropped  <- test_the_user_agent_is_passed_through
CAUGHT    launch args: DevTools host hard-coded  <- test_devtools_is_bound_to_the_given_loopback_host_and_port
CAUGHT    launch args: DevTools port ignored  <- test_devtools_is_bound_to_the_given_loopback_host_and_port
CAUGHT    launch args: --headless=new dropped  <- test_the_unconditional_flags_are_always_present
CAUGHT    launch args: --window-size dropped  <- test_the_unconditional_flags_are_always_present
CAUGHT    launch args: --log-level=3 dropped  <- test_the_unconditional_flags_are_always_present
CAUGHT    launch args: --disable-blink-features dropped  <- test_the_unconditional_flags_are_always_present
CAUGHT    launch args: --disable-logging dropped (the dedupe case's probe)  <- test_the_unconditional_flags_are_always_present
CAUGHT    launch args: base-arg de-duplication removed  <- test_a_base_argument_already_present_is_not_repeated
CAUGHT    launch args: bypass list never emitted  <- test_the_proxy_bypass_list_is_emitted_when_configured
CAUGHT    launch args: proxy flags emitted unconditionally  <- test_no_proxy_flags_are_emitted_when_the_proxy_is_not_configured

31/34 caught; survivors: ['sandbox: hasattr guard removed', "sandbox: negative spellings narrowed to '0'", 'retry attempts: .strip() removed']
```

</details>

The three survivors are the three the PR description names, and each is
explained in the docstring of the case that would have to kill it. None is a
gap:

- **`hasattr(os, "geteuid")` removed** — the bare call raises `AttributeError`
  where the attribute is absent, and the surrounding `except Exception` catches
  it identically. Behaviourally equivalent for every input.
- **Negative spelling tuple narrowed** — the resolver's fallthrough returns
  `False` too, so no input distinguishes "recognised as off" from "unrecognised,
  defaulted to off". That branch is dead code today; flagged, not deleted,
  because it stops being dead the day the default flips.
- **`.strip()` removed from the retry resolver** — `int` tolerates padding, and
  an all-whitespace value raises `ValueError` into the same fallback the
  stripped empty string reaches.

**On the unverifiable suite numbers.** Also fair — this repository has no CI job
that runs `pytest`, which is what a later step in this epic exists to fix, so
`13 failed → 7 failed` is a local measurement either way. The one part that is
independently checkable without running anything: the ledger's Linux block and
the suite's failing set are the same seven ids, and
`tests/test_baseline_failure_ledger.py` fails loudly if they ever diverge. Once
repo-wide CI lands, that guard is the check.

PR #60, opened 2026-09-01
Rebuild the browser-startup orchestration tests on autospec doubles (plan step E1-3)

shelpuk:

To read a JavaScript-heavy page, the server starts a headless Chromium in a
child process. Bringing one up is a sequence with several ways to go wrong: pick
a port, launch the browser, wait for it to answer, connect to it, retry if it
doesn't, kill any attempt that failed, delete the temporary profile. If the
retry or the kill misbehaves, browser processes accumulate on the machine and
are never cleaned up.

Three tests were supposed to guard that sequence. They had been failing since
the launch routine changed shape, and — as with the previous change in this
epic — repairing the obvious error would not have made them trustworthy. Four
further defects were hiding behind it, each measured rather than suspected.

They now pass, and they are hard to fool: an independently drawn set of 21
sabotages of the startup sequence is caught by 21 of them.

Suite: **7 failing → 4 failing**, 408 passing → 414.

---

## What this implements

Plan step **E1-3**, "Keep retry/cleanup orchestration at `_fetch_html`". These
claims stay at this layer permanently — `_fetch_html` is the child process's own
function and owns startup, retry and cleanup, and no resolver test can assert
that a process was terminated.

Three tests repaired in place, three added.

### The four defects behind the signature error

| Defect | Why repairing the signature alone would not have been enough |
|---|---|
| **Every double accepted any call.** Each collaborator was patched with a bare `AsyncMock`. | This is the original sin of the whole epic — it is exactly why five new required arguments disabled eight tests without one objecting. Every double is now `autospec`'d from the real callable. Demonstrated: adding a required argument to `_launch_chromium`, `_wait_for_devtools_ready`, `_terminate_process` or `_pick_free_port` now **fails** the tests. |
| **The fake `nodriver` module could never have worked.** | The tests installed `type("X", (), {"start": ...})` into `sys.modules`, but `_fetch_html` runs `cdp = uc.cdp` immediately after importing, before any branch. `nodriver` is an unconditional dependency, so the real module is used with only `start` patched — which is also what makes autospec of the connect call possible. |
| **A naive repair would have shelled out, invisibly.** | With `KINDLY_USER_AGENT` unset, `_resolve_user_agent` reaches `_detect_chrome_version`, which runs `<browser> --version`. Measured with a counting spy: **zero** real launches with the pin, **four** tests launching without it. A spy that *raises* sees nothing — `_detect_chrome_version` swallows it in `except Exception: pass`. |
| **The environment steered them.** | Fourteen variables now cleared per case, plus `shutil.which`, `os.geteuid` and the module's diagnostics global — none of the last three being environment variables. The module passes under a hostile environment and fails once the sweep is removed. |

### Two claims the step requires that no test made at all

- **A non-retryable error is not retried.** Nothing exercised the false branch of
  `_is_retryable_browser_connect_error`, so deleting the question turned every
  failure — a corrupt profile, a missing shared library — into three slow
  attempts and a message blaming the connection.
- **A missing browser executable surfaces actionable guidance.** This is the
  only useful message anyone gets when no browser is installed. It was the half
  of a test the previous step deleted that belonged to `_fetch_html` rather than
  to the resolver, and the plan was amended then to assign it here.

### A third, added during review

- **The retry backoff grows and scales for snap.** Every other case drives the
  base to zero, which makes `base * 2**attempt * multiplier` zero however it is
  written — so gutting the expression survived them all, while the harness
  docstring claimed the seam "exercises the real backoff arithmetic". One case
  now sets a real base and reads the durations back.

---

## The ledger drained under its strongest category

Running the guard **before** removing the ids reports all three under:

```
In the ledger, ran, and passed (a repair; delete the id):
  …::test_retries_and_terminates_on_devtools_timeout
  …::test_retries_on_failed_to_connect_to_browser
  …::test_uses_ignore_cleanup_errors_for_profile_dir
```

and raises no complaint of the other three kinds. That is the ledger's own rule
— *an id may leave only by passing* — met natively, because rewriting a test in
place keeps its node id. No relocation exception, no mapping table, nothing
taken on trust.

**This corrects a prediction the previous step recorded.** It warned that E1-3
and E1-4 would hit its relocation loophole. They do not, and the ledger now
explains why and re-aims the machine-readable-table recommendation at **E1-5**
by name — the one remaining step whose test name (`..._defaults_on_windows`)
cannot survive being made OS-neutral, so its id really will leave by deletion.

---

## A production defect found, reported, not fixed

`_is_snap_browser("/snap/bin/chromium")` returns **False** on Ubuntu. That path
is a symlink to `/usr/bin/snap`, and the function calls `os.path.realpath`
first — so the snap backoff multiplier does not apply to the single most common
way of having a snap-packaged Chromium. Measured on Ubuntu 24.04.

Fixing it is a production change with its own risk and is not this step's. The
backoff test therefore uses a fabricated `/snap/` path with no symlink to
follow. Worth noting that the *previous* version of this file used the real
`/snap/bin/chromium` believing it exercised the snap branch — it was asserting
the non-snap branch throughout.

---

## On the mutation evidence, honestly

The first battery here was 14 mutations, 14 caught. That claim was weaker than
it sounded: those mutations were written by the same author as the tests, from
the same mental model, so they agreed with them. An **independently drawn**
battery found **ten survivors**, the important one being `uc.start(port=port+1)`
— connecting on a different port than Chromium was launched on, which is
precisely the fault this retry loop exists to handle, green against every test.

Nothing asserted that the port and executable travelled intact between the four
collaborators, nor that a port is picked *per attempt* rather than once (reusing
one sends a retry at the port the previous, still-dying browser holds). Six
assertions close it.

The battery is now 21 and catches all 21 — stated as *coverage of what was
targeted*, not as absence of survivors.

<details><summary>21 mutations, 21 caught — full output</summary>

```
CAUGHT    R3: ignore_cleanup_errors dropped
            <- test_uses_ignore_cleanup_errors_for_profile_dir
CAUGHT    R3: profile prefix changed
            <- test_uses_ignore_cleanup_errors_for_profile_dir
CAUGHT    R1/R2: failed attempt is not terminated
            <- test_retries_and_terminates_on_devtools_timeout, test_retries_on_failed_to_connect_to_browser
CAUGHT    R1/R2: only one attempt is ever made
            <- test_backoff_grows_with_each_attempt_and_scales_for_snap, test_retries_and_terminates_on_devtools_timeout, test_retries_on_failed_to_connect_to_browser
CAUGHT    R4: the retryable check is dropped from the re-raise
            <- test_does_not_retry_a_non_retryable_error
CAUGHT    R5: the missing-browser raise is removed
            <- test_missing_browser_executable_names_the_override_variable
CAUGHT    R5: the override variable is dropped from the message
            <- test_missing_browser_executable_names_the_override_variable
CAUGHT    R1: the loop never breaks on a successful connect
            <- test_retries_on_failed_to_connect_to_browser
CAUGHT    R1: final cleanup no longer terminates the process
            <- test_retries_on_failed_to_connect_to_browser
CAUGHT    R2: the attempts-exhausted diagnosis is dropped
            <- test_retries_and_terminates_on_devtools_timeout
CAUGHT    R1: the browser is never stopped
            <- test_retries_on_failed_to_connect_to_browser
CAUGHT    R6 autospec: _launch_chromium gains a required argument
            <- test_backoff_grows_with_each_attempt_and_scales_for_snap, test_does_not_retry_a_non_retryable_error, test_retries_and_terminates_on_devtools_timeout, test_retries_on_failed_to_connect_to_browser, test_uses_ignore_cleanup_errors_for_profile_dir
CAUGHT    R6 autospec: _terminate_process gains a required argument
            <- test_backoff_grows_with_each_attempt_and_scales_for_snap, test_does_not_retry_a_non_retryable_error, test_retries_and_terminates_on_devtools_timeout, test_retries_on_failed_to_connect_to_browser
CAUGHT    R6 autospec: _wait_for_devtools_ready gains a required argument
            <- test_backoff_grows_with_each_attempt_and_scales_for_snap, test_devtools_probe_ignores_proxy_env, test_does_not_retry_a_non_retryable_error, test_retries_and_terminates_on_devtools_timeout, test_retries_on_failed_to_connect_to_browser, test_uses_ignore_cleanup_errors_for_profile_dir
CAUGHT    W2: readiness probes a different port than was launched
            <- test_uses_ignore_cleanup_errors_for_profile_dir
CAUGHT    W2: connect targets a different port than was launched
            <- test_uses_ignore_cleanup_errors_for_profile_dir
CAUGHT    W2: a different executable is launched than was resolved
            <- test_uses_ignore_cleanup_errors_for_profile_dir
CAUGHT    W2: the port is picked once instead of per attempt
            <- test_retries_and_terminates_on_devtools_timeout
CAUGHT    W3: the backoff exponent is dropped
            <- test_backoff_grows_with_each_attempt_and_scales_for_snap
CAUGHT    W3: the snap multiplier is dropped from the backoff
            <- test_backoff_grows_with_each_attempt_and_scales_for_snap
CAUGHT    W3: the profile-flush wait is removed
            <- test_backoff_grows_with_each_attempt_and_scales_for_snap

21/21 caught; survivors: none
```

</details>

---

## A timeout that took five times its own budget to unwind

The review found the per-test bound was necessary and not sufficient. Measured:
with the retry loop mutated to spin forever, the 10 s `asyncio.wait_for` fired
and the test then took **47 s** to report — because every doubled call is
recorded, and a loop with zero backoff records millions of them. A timeout that
behaves that way is the outcome a timeout exists to prevent.

Fixed at the cause rather than the symptom: the launch double refuses an
eleventh browser and fails with a sentence naming the loop. **47 s → 1.4 s.**
The wall-clock bound drops to 5 s and now only catches a hang that makes no
progress at all. Both facts are recorded in §5.4 so the next test at this layer
does not rediscover them.

---

## One decision deliberately not made

The design routes this group to the `subsystem` job, whose registered meaning is
*"needs a real socket or child process"* — which these tests, written correctly,
do not. Unmarked they fall into `fast`, and that matters beyond routing:
`.coveragerc-gate` omits `nodriver_worker.py`, and control 1 requires every
omitted module to show non-zero coverage in the observational report, which
`fast` does not produce.

Three resolutions are named in §5.2 for **E4-1**, which has to answer this
anyway. E1-3 changed nothing: the marker list is a contract guarded against the
design document, the control that depends on the answer does not exist yet, and
widening a registered meaning on speculation is the wrong direction.

---

## Verification

| Check | Result |
|---|---|
| Full suite | 4 failed, 414 passed, 2 skipped (from 7 failed, 408 passed) |
| Failing set vs. the ledger's Linux block | identical |
| Failing-set diff vs. `main` | three removals, **zero additions** |
| This module | 8 passed, 1.2 s |
| Mutation battery | 21 targeted, 21 caught |
| Real subprocess launches (counting spy) | **0** — and 4 tests launch once the user-agent pin is removed |
| Hostile environment exported | passes; fails once the sweep is removed |
| Ledger guard, incl. live child run | 15 passed |
| Review system's own suite | 482 passed |
| `scripts/check_plan_dag.py` | exit 0 |
| `ruff check` on the changed file | clean |

Reviewed by `system-design-reviewer` and `code-reviewer`; both initially
requested changes and every finding is implemented or recorded in the design
with its owner.

### DIFF
--- a/.system_design/BASELINE_FAILURES.md
+++ b/.system_design/BASELINE_FAILURES.md
@@ -50,13 +50,30 @@ one is a dropped claim, whatever the pull request calls it.
 E1-2 evidenced the table by hand, running the guard in the intermediate state —
 tests deleted, ledger not yet drained — so that it named those five ids as *no
 longer collected* and named nothing else. That is a real proof and it is also a
-one-off: nothing re-runs it. E1-3 and E1-4 rewrite their tests rather than
-relocate them, so their ids change too and they hit the same wall. **Whichever
-of them lands first should make this table machine-readable** — a fenced block
-of `<deleted id> -> <replacement id>` — and assert that every replacement is in
-the child run's *collected* set and absent from its *failed* set. The guard
-already holds both sets; the cost is one assertion, and it converts this
-section from a promise into a check.
+one-off: nothing re-runs it.
+
+**E1-2 predicted that E1-3 and E1-4 would hit the same wall. They do not, and
+the reason is worth keeping.** A step that *rewrites a test in place* keeps its
+node id, so the id leaves this ledger by passing and the rule holds natively
+with no exception and nothing to take on trust. E1-3 rewrote three test bodies
+under their existing names and the guard reported all three under *"In the
+ledger, ran, and passed"* — the strongest of its four categories, and the one
+that needs no supporting document. E1-4 replaces a fixture, so it is in the same
+position.
+
+**The step to watch is E1-5**, which rewrites the concurrency test to be
+OS-neutral. Its id is `test_web_search_concurrency_defaults_on_windows`, and a
+name that still says `on_windows` would be wrong afterwards — so that step
+probably *will* rename, and its id will leave by deletion. It is the remaining
+candidate for making this table machine-readable: a fenced block of
+`<deleted id> -> <replacement id>`, asserted to be in the child run's *collected*
+set and absent from its *failed* set. The guard already holds both sets. If
+E1-5 keeps the name instead, no such machinery is needed at all.
+
+**The rule that generalises: prefer rewriting in place to relocating.** Not for
+tidiness — it is the difference between an id that leaves this ledger under a
+machine-checked category and one that leaves under a promise in a pull request.
+Relocate only when the layer itself is the defect, as it was for E1-2.
 
 Adding an id here is not a way to make a new failure acceptable. A new red test
 is a regression; this ledger names what is left of the pre-existing twelve, and
@@ -179,7 +196,7 @@ an empty failing set on a red suite, and two of this repository's three
 `subTest` sites are in `tests/test_universal_html_loader.py`, one of the files
 this ledger tracks.
 
-## What the remaining seven are
+## What the remaining four are
 
 Grouped by cause, so a reviewer can tell which repair step owns which id. The
 grouping is by module rather than per id, because the blocks are sorted and so
@@ -188,9 +205,12 @@ from it.
 
 | Module | Ids | Cause | Repaired by |
 |---|---|---|---|
-| `test_nodriver_worker_sandbox.py` | 3 | `_fetch_html` grew five required keyword-only arguments; the callers were never updated | E1-3 (retry and cleanup). Its other five ids were the flag-and-default half and left with E1-2 |
 | `test_universal_html_loader.py` | 3 | `fetch_html_via_nodriver` streams `proc.stdout`; `_FakeProc` never grew one | E1-4 |
 | `test_server.py` | 1 | asserts a Windows concurrency cap that `_resolve_web_search_max_concurrency` no longer has | E1-5 |
+
+`test_nodriver_worker_sandbox.py` has left this table entirely: its five
+flag-and-default ids relocated with E1-2 and its three orchestration ids were
+repaired in place by E1-3.
 
 ## Why the two platforms differ
 
@@ -226,7 +246,7 @@ claim awaiting its lane.
   Linux 6.8.0-138 · CPython 3.13.15 · pytest 9.1.1 · pytest-asyncio 1.4.0 ·
   mcp 1.29.1 · starlette 1.6.0 · uvicorn 0.52.4 · nodriver 0.50.3
 - **Result:** 12 failed, 303 passed, 2 skipped, 9 subtests passed
-- **Remaining:** 7 failed
+- **Remaining:** 4 failed
 
 This is the figure §1.1 predicted from source and labelled unmeasured. It was
 measured, and it matched: twelve, being the eight stale `_fetch_html` callers,
@@ -237,9 +257,6 @@ seven remaining, listed below. The **Result** line keeps saying twelve because
 that is what the run said.
 
 ```text
-tests/test_nodriver_worker_sandbox.py::TestNodriverWorkerSandbox::test_retries_and_terminates_on_devtools_timeout
-tests/test_nodriver_worker_sandbox.py::TestNodriverWorkerSandbox::test_retries_on_failed_to_connect_to_browser
-tests/test_nodriver_worker_sandbox.py::TestNodriverWorkerSandbox::test_uses_ignore_cleanup_errors_for_profile_dir
 tests/test_server.py::TestWebSearchTool::test_web_search_concurrency_defaults_on_windows
 tests/test_universal_html_loader.py::TestUniversalHtmlLoader::test_fetch_html_passes_browser_executable_path_when_set
 tests/test_universal_html_loader.py::TestUniversalHtmlLoader::test_fetch_html_sets_no_proxy_for_loopback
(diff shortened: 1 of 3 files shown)


shelpuk:
## Review addressed

**The one suggestion is fixed** in `aa64772`, and it was right. Verified before
editing: `KINDLY_HTML_TOTAL_TIMEOUT_SECONDS` is read only by
`_resolve_worker_timeout_details` and `KINDLY_DIAGNOSTICS` only by
`_diagnostics_enabled`, and `grep` confirms both have exactly one caller —
`_main_async`, the worker's entry point. Neither is reachable from
`_fetch_html`.

The harness was behaving correctly (over-clearing errs safe); the *claim* of
equivalence was wrong. The comment now says **superset**, names both entries, and
records why they are cleared anyway: a future change that routes either into
`_fetch_html` should not quietly make these cases steerable.

## On the two claims the review flagged as unverifiable

Both are fair. Filling in what can be shown from the repository rather than
asking for trust.

**The Ubuntu symlink**, which the review rated medium-confidence:

```console
$ python -c "import os; print(os.path.realpath('/snap/bin/chromium'))"
/usr/bin/snap
$ python -c "from ...nodriver_worker import _is_snap_browser as f; print(f('/snap/bin/chromium'), f('/snap/bin/chromium-for-tests'))"
False True
```

Measured on Ubuntu 24.04.4. The code-level reasoning the review reached
independently is the whole of it: `os.path.realpath` runs before the `/snap/`
test, and `/usr/bin/snap` contains no `/snap/` substring. Anyone can reproduce it
on a machine with snap Chromium installed.

**The 21-mutation battery**, which the review correctly notes is claimed and not
demonstrated in the diff. Its full output is in the PR description, and the
important half is not the count — it is that ten of an *independently drawn*
battery survived an earlier revision of this work, including a connect call
aimed at a different port than the browser was launched on. My own mutations
came from the same mental model as my own tests and agreed with them. The claim
is stated as coverage of what was targeted, not as absence of survivors.

**On not running tests.** Noted, and it is not this review's limitation to fix:
the repository has no CI job that runs `pytest` at all, so every test figure on
every PR here is a local measurement. Closing that is E4's job, gated on E1-6 —
two steps away now. The nearest independent check available today is that the
ledger's Linux block and the suite's failing set are the same four ids, and
`tests/test_baseline_failure_ledger.py` goes red the moment they diverge.

PR #61, opened 2026-09-01
Declare the worker-process contract, and prove the check can fail (plan step E2-1)

shelpuk:

The server fetches awkward web pages by launching a **separate browser process** and reading its output. Nothing in the codebase wrote down what that process *looks like* — which pieces the main program reads off it. That contract existed only as an unwritten understanding between the code that starts the process and the stand-ins used in tests.

**It already failed once, and the damage is still on the board.** The main program was changed to read the process's output stream directly. The stand-in used by three tests was never updated to match. Those three tests have been broken ever since, and nobody noticed at the time because nothing was checking.

This PR writes that contract down once, in one authoritative place, and puts guards on it so the same drift cannot happen silently again.

It deliberately **adds no behaviour and changes no production code path**. Repairing the three broken tests is the next ticket; applying this contract to production code is the one after. This is the foundation both of those stand on.

### Why three guards and not one

Each mechanism misses something the others catch. This is measured, not assumed:

| Guard | Catches a *missing* piece | Catches a *changed* signature |
|---|---|---|
| Runtime presence check | yes | **no — passes silently** |
| Static type check | yes | yes |
| Real-process test (later ticket) | behaviour only | behaviour only |

That middle-column gap is the exact shape of the original outage. A runtime check reports "all good" while the contract is broken.

---

## Four things the design assumed that turned out not to hold

Each was found by measurement during implementation, and each would have shipped a check that passes while proving nothing. This is the substance of the PR.

**1. The documented defence against a hollow "mock" object does not work.**
The design said a particular strict setting stops someone swapping the real stand-in for a hollow mock that satisfies any contract at all. Measured: it does not fire — the mock isn't "untyped", it's a concrete type whose *members* are untyped, and the setting only looks at the former. I falsified my own test setup first to be sure the tooling was live. Replaced with a mechanism I measured working, and froze the trap as a permanent fixture. **Section 8A step 3 of the design document said otherwise and is corrected here.**

**2. The obvious spelling of the contract would have blocked the next ticket.**
Written the natural way, the contract is one the **real** browser process fails to satisfy. The follow-up ticket passes a real process to code annotated with it, so that would have surfaced two tickets downstream as a mystery. A different spelling accepts both the real process and the stand-in. The trade-off it costs is recorded in the code, so nobody "simplifies" it back.

**3. The type checker would have silently given up.**
This project doesn't publish type information about itself, so run naively the checker skips the contract entirely and treats it as "anything goes" — under which **every** conformance check passes. The green light would have been meaningless. Every check now also asserts the checker actually resolved the code.

**4. The deliberately-broken fixtures were not broken.**
The fixtures that exist to prove the check *can* fail would have type-checked clean, making the harness red from day one for the wrong reason. Each now declares its own strictness where a reader meets it, and the harness asserts a **specific** diagnostic — not merely "it failed", since the tool also fails on a crash or a typo.

---

## Settling an open question the design left to this step

The design document explicitly deferred one decision here: whether the type-check harness belongs in the fast, hermetic test lane. It does not — it launches a child process. So it is marked accordingly, and the type checker was removed from the pinned dependency set that lane installs, in all four places that had to agree.

The removal was **computed, not guessed**: a clean-room reinstall constrained by the lockfile's own previous contents differed by exactly the type checker and its exclusive dependencies, and nothing else. No unrelated version churn rides along in this PR.

**One test disappears, and that is intended.** `test_extra_declares_expected_bound[ratchet-mypy]` is generated from the table this PR edits, so removing the entry removes its case. The claim is not dropped — the tool stays asserted by two other cases, and the set is still pinned in both directions. Flagging it because a vanishing test and a dropped claim look identical in a diff.

---

## Evidence

- **Failing-test set: identical to `main`.** Not the count — the set. The four pre-existing failures remain and belong to the next two tickets. A count comparison cannot tell a fix from a fix beside a new break.
- **Every guard was falsified before being trusted.** Six deliberate breakages, each watched going red for the right reason: delete the exclusion; mistype a config key; break the source path; remove a member from the stand-in; defuse each of two fixtures.
- **One failure mode found by review and fixed:** without a guard, removing a member from the stand-in aborted collection of the *entire suite* — reporting one defect as "0 tests ran". Now it fails as the single case that owns the claim, with the suite still running.
- 24 new tests, `mypy` clean, `ruff` clean on changed files, dependency-graph check green, review-system suite green (482).

## Reviewing this

The highest-value question is **"could this test pass for the wrong reason?"** — that is what the whole change exists to prevent, and it is where the four findings above came from. `tests/typing_negative/` is the place to look: each fixture is a deliberate defect, and each records what it proves and why the obvious alternative does not.


shelpuk:
## Findings addressed — pushed as `15f6ff1`

Three reviews (design, code, and the automatic reviewer here) converged on the same drift, so it was real. All findings actioned below; every claim re-measured rather than taken on trust.

### Warnings — all fixed

| Finding | Action |
|---|---|
| `TEST_SUITE.md:828` undercounts the fixture set (3 vs 4) | Corrected to four, **with the reason the fourth exists** |
| `python-tests.md:64` rule row says "three" | Corrected, plus the `strict_bytes` clause |
| "every fixture declares its strictness inline" | Corrected — only **one of four** does |

The fixture-count drift mattered more than a typo: the rule row is what tells this reviewer which guards may not be weakened, so a PR **deleting the `bytearray` fixture would have read as compliant with the rule as written** — and that fixture is the only thing enforcing the `strict_bytes` constraint the plan assigns to this step.

The inline-strictness claim had a sharper edge. The `exclude` comment in `pyproject.toml` ended *"which is why they need their own override below"*. **There is no override below, and adding one is what the design forbids** — it would match nothing, `warn_unused_configs` would report it, and `test_ordinary_target_reports_no_unused_configuration_sections` would go red. That comment instructed the next author to break a guard. Rewritten to say so explicitly.

### Suggestions — all fixed

- **`TEST_SUITE.md:759` / `types.py:47` arithmetic.** Correct: four of the seven members carry state, the other three are methods. "Every member is a read-only property" and "the other six" were both wrong. Fixed in both places.
- **`bytearray_payload.py:30` — no inverse-proves-it guard.** The best finding of the three suggestions. The sibling vacuity check strips a fixture's own inline directive; this fixture rests on a mypy *default* instead, so the equivalent proof is to invert the default. Added `test_the_bytearray_fixture_is_clean_without_strict_bytes`, which asserts it passes under `--no-strict-bytes` — the mypy 1.x behaviour this project deliberately skipped. Verified by hand before; automated now.

### One finding that does not reproduce

`tests/typing_negative/__init__.py:26` — *"\_\_init\_\_.py claims every fixture declares its strictness inline"*. **That file is 13 lines long and makes no such claim.** Both fixture docstrings that mention inline strictness are correctly scoped: `any_typed_double.py` says *"declared inline, **here**"* about itself, and `double_missing_stdout.py` says *"It needs no inline strictness."* Left unchanged. The underlying claim was real, but it lived in `TEST_SUITE.md` and `pyproject.toml` — both now fixed.

### Also closed, from the design review

- **A vacuity in the mypy-absent case.** mypy writes to stdout whenever it runs at all, so silence plus a non-zero exit means it never started. `test_negative_fixtures_are_absent_from_a_whole_tree_run` asserts an *absence* in stdout — which silence satisfies. `_run_mypy` now raises on that condition, naming the install as the cause, instead of surfacing as blank assertion messages.
- **§10.3's job table gains an Install column.** Removing mypy from the pinned lane made "which extra does this job install" load-bearing for the first time, and three future job definitions depend on it — not the one the earlier note named.
- **Two landmines recorded for the follow-up step**, which is sized on a narrower reading: two functions already take a concrete process type, and `universal_html.py:185` already fails the type check today (`STARTUPINFO` — mypy narrows on `sys.platform`, not on the `os.name` guard the code uses).

### Verification after the changes

- Failing-test **set** still identical to `main` — the four pre-existing failures, owned by the next two tickets.
- 25 tests · mypy clean · ruff clean on changed files · review-system suite 482 passed · dependency-graph check exit 0.
- Falsification re-run after every edit.

On the note that the fourth fixture wasn't in the PR description: correct, and that omission is what produced the drift. Fixed at the source.

PR #62, opened 2026-09-02
Repair the three loader tests on the canonical typed fake (plan step E1-4)

shelpuk:

The server fetches awkward web pages by launching a **separate browser process** and reading its output. Three tests check that the parent launches that process *correctly* — right command line, right environment.

Those three have been **broken since production changed how it reads the child's output**. Each test carried its own private copy of a stand-in for the browser process; production moved on and the copies didn't. Three copies of the same thing, none of them tied to what production actually does — which is the exact condition that caused the breakage.

This PR points all three at the **single canonical stand-in** shipped by the previous ticket, so there is one definition instead of four.

**No production code changes.** The failing count goes **4 → 1**; the one remaining failure belongs to the next ticket.

### Why these tests need a stand-in at all

They verify what the parent *passes* to the child. Once the child is running that information is gone — a real browser cannot tell you what command line produced it. This is the textbook case for a test double, and the design says so.

---

## Two defects found beyond the missing attribute

Both were surfaced by review, not by my own first verification pass. Both are more serious than the original ticket.

### 1. The tests could pass while asserting nothing

With the production behaviour **completely deleted**, two of the three still pass — on any machine that happens to export a common environment variable:

| Environment | Result with production behaviour removed |
|---|---|
| Clean shell | Fails correctly ✅ |
| Exports `PYTHONPATH` / `NO_PROXY` | **Passes — proving nothing** ❌ |

My first falsification pass ran six sabotages and reported six catches. That was **luck**: my shell exported neither variable. On most developer machines and CI images, two of these tests would have been decorative.

Fixed by clearing a declared superset of every variable the loader reads, then having each test declare what it needs. All six sabotages re-run under both a stripped-bare and a deliberately hostile environment — six for six, in both.

### 2. The tests were leaking real browsers

They were entering the browser-pooling path and only reaching the code under test because pooling *failed* and the error was swallowed. Measured: the pool was entered once per test, leaving a populated module global, an `atexit` hook, and **live headless Chrome process trees with `/tmp` profile directories still running afterwards**.

Now pinned off — measured **zero** pool entries. The pooled command variant that consequently loses coverage is handed to the next ticket in writing rather than left to be discovered.

---

## One new guard, and an honest account of its limits

"Use the one canonical stand-in" was enforced by nothing but a human eyeball. This adds a check that fails if any test file reimplements that shape.

**Its reach is narrower than it first appears, and I initially described it wrongly.** My first version of the failure message claimed it "is what left three loader tests red." That is **false** — the double that caused the outage named only two members, and this guard would not catch it. What it catches is a *whole-surface* reimplementation: the shape a future author writes when they rebuild the double instead of importing it.

Corrected in the guard's message and in both design documents. The limit is pinned by a pair of synthetic cases so nobody has to rediscover it, and breaking the detector now fails a test rather than passing quietly.

---

## Design-document corrections this step owed

- **§11.2 rules out the very seam these tests use.** It survives here as a *stated interim*: the drift half of that objection is closed, the coupling half stands until the runner is extracted. Now recorded at both ends rather than inferable only from a diff.
- **§8 A.3 predicted this step would introduce a problematic double.** It didn't — so a stale prediction became a stated invariant.
- **Two ledger paragraphs still said "four remaining"** above a block listing one.

## Ledger

Drained by exactly three IDs on both platform blocks; the frozen measurement lines are untouched. The IDs leave **by passing**, under their original names — the guard's strongest category, no exception owed.

The **Windows** half is drained by argument, not measurement: no Windows lane exists yet. These tests never spawn a real process and touch no OS-specific branch, so the repair is platform-neutral — but that is an argument, and it is labelled as one.

## Evidence

- Failing **set** = previous set minus exactly the three IDs. The set, not the count — a count cannot tell a repair from a repair beside a new break.
- Six sabotages of production, each caught by exactly one test, in **two** environments. One of them reproduces the original outage.
- Ledger guard (re-runs the suite in a child process and compares IDs) passes. Type checker clean. Plan graph green. All 9 lint findings in the touched file are pre-existing and unchanged.

## Reviewing this

The highest-value question is **"could this test pass for the wrong reason?"** — that is where both real defects came from. The environment-clearing helper and the guard's `DETECTOR_CASES` are where the answers live.


shelpuk:
## Suggestion addressed — pushed as `54202c3`

Thanks — the point underneath it is fair, though the premise is slightly off and worth correcting for the record.

**The Windows argument is not inside the Linux section.** It sits in *"What the remaining one is"* (ends line 244), which is platform-neutral and above **both** blocks; the Linux section starts at line 273 and Windows at 295. So the argument is not misfiled.

**What is true, and worth fixing:** a reader landing on the Windows block sees three ids gone from a platform nobody ran, with no local explanation and no pointer to where one exists. That is exactly the kind of gap this document's discipline exists to close.

Added a pointer in the Windows block, stating plainly that E1-4's three left **by inference rather than by a run**, and naming what the inference rests on — the three loader tests spawn no process and touch no `os.name`/`sys.platform` branch. E1-2's four and E1-3's three are in the same position for the same reason, which the block now says.

The argument itself stays stated **once**, where it was. Repeating it in both places would be two copies of a claim that can drift — the failure mode this ledger already carries scar tissue about.

Ledger guard re-run: 15 passed. Suite unchanged at `1 failed, 444 passed`.

PR #63, opened 2026-09-02
Rewrite the concurrency test OS-neutral, and drain the ledger (plan step E1-5)

shelpuk:

**The test suite is green for the first time in this workstream.** `0 failed, 448
passed, 2 skipped`. That was the whole point of this epic, and it is the gate the
CI epic waits on: today a green checkmark on a pull request says nothing about
this repository's own tests, because CI deliberately does not run them while they
are red. This is the last repair; the milestone step that follows has no diff.

**Nothing user-facing changes and no production code is touched.** The server has
a knob, `KINDLY_WEB_SEARCH_MAX_CONCURRENCY`, that decides how many result pages
are fetched in parallel. There used to be a rule capping Windows to one at a
time. That rule was removed from the product some time ago; one test still
asserted it, by faking the operating system, and had been failing ever since.

**Two things worth knowing beyond the ticket:**

1. **The "at most 5 parallel fetches" ceiling was protected by no test at all.**
   The one case that passed a larger number also passed a small result count, so
   a second, unrelated limit produced the right answer for the wrong reason.
   Measured, not assumed: deleting the ceiling from the product code on the old
   test file changed nothing. This matters commercially — the ceiling is what
   stops one search from opening an unbounded number of browser tabs, and we ship
   the promise `clamped 1..5` in the tool description the calling AI model reads.
   It now has a test.
2. **The ticket's own acceptance criterion could not be satisfied as written**,
   for that same reason. Rather than pass an unmeetable check quietly, the plan
   is corrected and the reason recorded, so the next reader does not score this
   step incomplete.

---

## What changed

Three methods in `tests/test_server.py` patched
`kindly_web_search_mcp_server.server.os.name` and split their expectations by the
patched value. `_resolve_web_search_max_concurrency` has no platform branch, so
the split was fiction: one method failed on four of its five cases, and the other
two passed only by coincidence of their inputs. All three become one table-driven
test of seven rows.

Each row carries the single-line mutation of the resolver it detects, in
`ConcurrencyCase.pins`, reported in the failure message when the row fails. The
mapping sits on the row rather than in prose so it cannot drift from the case it
describes, and so a red build tells the reader which production line the row was
watching.

## Evidence

| Check | Result |
|---|---|
| Falsification battery | 7 single-line mutations × 2 environments = 14 runs, each killing **exactly** the predicted rows, plus 2 unmutated controls |
| Ceiling was uncovered | measured on the parent commit — deleting `min(value, 5)` left the old failing set unchanged |
| Intermediate ledger state | guard named exactly one id under *"In the ledger but no longer collected"*, nothing under the other three categories |
| New guard falsified | 4 ways — bad replacement id, a test resurrected under a retired id, a `Result` bullet added to the new section, a malformed row |
| One child run | both slow tests together 20.4s; the new one alone 20.1s — the second consumed a cached fixture |
| `-m "not slow"` | 15 passed, 2 deselected, 0.35s — no child spawned |
| Full suite | **0 failed, 448 passed, 2 skipped, 16 subtests** |
| `mypy` (no args) | clean |
| `ruff check` | clean on the guard; `test_server.py` went 8 findings → 6, none new |

The battery runs under **two** environments — one stripped bare, one deliberately
exporting `KINDLY_WEB_SEARCH_MAX_CONCURRENCY` — because a control that only fires
on one machine is no control. This repository paid for that lesson one step ago.

The positivity guard needs **two** mutations, not one: `0` is already falsy, so
dropping the `> 0` conjunct kills the negative row and leaves the zero row
passing. An earlier draft mapped both rows to a single mutation and would have
shown a legitimate row surviving, inviting its deletion.

## The ledger, and why it grew a second invariant

This repair **renames**, so its node id leaves by deletion rather than by
passing. That is the ledger's known loophole: its "listed but no longer
collected" complaint fires only for ids *still listed*, so deleting a test and
its ledger entry in one change is indistinguishable from repairing it.

`BASELINE_FAILURES.md` now carries a machine-readable `## Relocated claims` block
of `<retired id> -> <replacement id>` rows, and
`tests/test_baseline_failure_ledger.py` holds each replacement to the same child
run it already makes: collected, and in neither the failed nor the skipped set,
with the retired id absent. **All three retirements are listed**, not only the
one the ledger named — the two that passed cost one line each and are now
machine-checked facts rather than sentences in this description.

Two decisions are declined **in writing**, in the document, with reasons:

- The five older relocation rows are **not** retrofitted. They map to functions
  rather than node ids, several replacements are parameterized, and one claim was
  split across two steps — so asserting that mapping now would attribute to
  another step's run a fact that run did not produce, which is precisely the
  failure the ledger's `Result:`/`Remaining:` split exists to prevent.
- Rows get a **retirement rule**, because E5-3 owns exhaustive coverage of this
  same resolver and may rename the replacement. Without a stated exit this guard
  would go red for a correct change, and the only apparent remedy would be
  weakening a guard.

Both platform blocks are drained; both `Result:` lines are untouched. Linux is
measured. **Windows is an argument, and it is now stated precisely**: no repair
*asserts a value computed by* a branch reading `os.name` or `sys.platform`. The
looser "touches no such branch" would be false — the loader tests' spawn path
does execute `os.name != "nt"` — and E1-6 must still take its Windows figure from
a real run, which that section says.

## Handed forward, and one loss declared

Three gaps go to E5-3 in writing: `num_results <= 0` (which skips the result-count
bound and would otherwise build an `asyncio.Semaphore(0)` that never opens), a
blank value, and the unreachable `max(1, ...)`.

One coverage loss is declared in three places — the test docstring, the ledger
and `TEST_SUITE.md`: the three deleted methods were the only thing in the tree
that would notice an `os.name` branch reappearing in this resolver. Nothing
asserts its absence now; the milestone's one real Windows run is what covers it
next.

## Review

Reviewed by `system-design-reviewer` (three blockers, all resolved — a mutation
mapping that was empirically wrong, an unowned Windows-drain argument, and a
substitution table that did not do what the design asked) and by `code-reviewer`
(**APPROVE**; its two warnings and four suggestions are all applied, including
lifting the new parser rejections into synthetic-document cases so each stays a
test rather than a one-time observation).


shelpuk:
Both findings addressed in 8fb49ec. No production code and no test behaviour changed.

**`tests/test_server.py:149` — the docstring claim was overgeneralized.** Correct, and it was falsifiable against this step's own battery. The docstring read *"the positivity guard needs two mutations because neither alone kills both of its rows."* Measured: replacing the condition with `if parsed is not None` kills **both** the zero and the negative row. Only the other mutation — dropping the `> 0` conjunct, leaving `if parsed` — is asymmetric, and it kills the negative row alone because `0` is already falsy.

So the reason the second mutation must exist by name is not that neither suffices. It is that a battery running only *that* one would report the zero row as dead weight and invite its deletion. The docstring now says that. Worth noting where the error was: the commit message, this PR body and `TEST_SUITE.md` §A.4 all carried the precise wording; the docstring was the single place that generalised it — and the place a future author reads first.

**`.system_design/TEST_SUITE.md:68` — "declared in three places" was true of two.** Also correct. The coverage loss (nothing now notices an `os.name` branch reappearing in this resolver) was stated in the test docstring and in the ledger, not in `TEST_SUITE.md`. I added it to §A.4 rather than weakening the claim: a coverage loss recorded only where the deletion happened is found only by whoever goes looking there.

Suite unchanged: `0 failed, 448 passed, 2 skipped, 16 subtests`.

shelpuk:
Addressed in 20f7fb6.

**`tests/test_baseline_failure_ledger.py:336` — the parenthetical was inaccurate.** Correct, and the inaccuracy was of a specific kind worth naming: I borrowed the ledger env sweep's language ("deliberately over-collecting shape scan") and applied it to a scanner whose own comment says the opposite.

`UNITTEST_FRAMEWORK` anchors its two *import* alternatives to line start and states that this is deliberate — *"Matched at line start so the string quoted inside a test is not evidence"* — while the third alternative, the one naming the two case classes, carries no such anchor and so matches inside prose. That is a **gap between the comment and the pattern**, not a design choice, and calling it deliberate would have told the next reader not to bother fixing it.

The behaviour the note warns about is unchanged, and was measured during this step rather than reasoned: a Sphinx cross-reference to that dotted name, in a *docstring*, makes the module report as unittest-style and claimed by no migration batch, failing `tests/test_plan_dag.py`. Found because the ledger guard went red — `scripts/check_plan_dag.py`, run directly, had passed minutes earlier in the same session.

Still not this step's script to change; flagging it for whoever owns that validator.

Suite unchanged: `0 failed, 448 passed, 2 skipped, 16 subtests`.

shelpuk:
Addressed in c6dff26 — and measuring it produced a sharper statement than the prose it replaces.

You are right that the sentence was still inaccurate, though the reason is one step further in. I ran the shipped `UNITTEST_FRAMEWORK` pattern against each candidate spelling rather than reasoning about it again:

```
unittest.TestCase.subTest   trips
TestCase.subTest            clean
unittest.TestCase           trips
```

So **the method is incidental**. The pattern matches the `unittest.<case class>` prefix that any *module-qualified* spelling carries — the class-qualified form is perfectly fine. "That method's dotted name would trip it" was wrong in the same way for a third time; the note now names the module prefix as the cause.

I also added the check I should have run first: the file is now verified against the shipped pattern itself, which reports **zero matches**.

The rest of the note stands, as you say: the alternative naming the case classes carries no line-start anchor, while the pattern's own comment claims that anchoring is deliberate *"so the string quoted inside a test is not evidence."* A gap between comment and pattern — still not this step's script to change, and still flagged for whoever owns that validator.

**On the previous round's red `review` check:** that was the 15-minute job cap, not a finding. The model step was cancelled at `15m20s` and every downstream step (`Interpret result`, `Post review`, `Fail when no review was produced`) was skipped, so no review was produced and the job went red by design. Re-ran it; it passed at `14m24s` — 50 seconds of headroom. The two passes before it took `11m27s` and `7m16s`. Worth a `timeout-minutes` bump for whoever owns the workflow, since a red check there is indistinguishable from a code problem at a glance.

Suite unchanged: `0 failed, 448 passed, 2 skipped, 16 subtests`.

PR #66, opened 2026-09-02
Move the worker runner out of the loader, and close the pool's release gaps (plan step E2-3)

shelpuk:

When a page needs a real browser to render, this server launches headless
Chromium as a **separate process** and reads the rendered page back over a pipe.
Until this PR, the code that babysits that process — starting it, draining its
output, keeping it alive, killing it when it hangs — lived in the same file as
the code that decides how to *parse* a page. About 900 lines of process plumbing
wrapped around a well-tested parsing path.

That mixture was blocking three things on the roadmap, because the quality gate
we are building can only include or exclude a **whole file**. Include that file
and every ordinary change to the browser plumbing fails a check it has no way to
satisfy; exclude it and the parsing path loses real coverage enforcement. Both
answers were wrong, so the file had to be split. This PR splits it.

**It also fixes four real defects found in the open code, two of which were not
on the ticket.** The one worth knowing about: under a specific failure sequence
the browser pool could hand **the same browser to two concurrent requests** —
same browser, same profile directory, two callers. Nobody had spotted it. The
others are a browser slot that leaked out of the pool on early failure
(degrading the server toward cold, slow browser launches), a retry path no test
could see, and a dead assignment.

No user-facing behaviour changes. Same inputs, same outputs, same diagnostics in
the same order — that last one is now asserted rather than assumed.

---

## What changed

**New module `scrape/worker_runner.py`** holds the spawn, both stream readers,
the output accumulators, the heartbeat, the process-tree terminator, the
diagnostics pipe probe, and a new
`_run_worker_command(command, *, env, default_timeout_seconds, diagnostics)`.
`universal_html.py` keeps the Markdown-probe path, the command builder, the
environment helpers and the pool orchestration.

The helpers **moved verbatim**, existing lint findings included. An extraction
that also rewrites what it moves cannot be reviewed as an extraction.

`.coveragerc-gate` has named `*/scrape/worker_runner.py` in its `omit` list
since before the file existed. It now matches something, and the guard that
reads that list asserts the runner's absence from the gating report for the
first time.

**The boundary is asserted, not intended.** `universal_html.py` imports neither
`asyncio` nor `subprocess` — both are required to start, wait on, stream from or
kill a process, so their joint absence *is* the "no process management here"
claim rather than a proxy for it — while still defining the command builder and
the probe path, so the split cannot be satisfied by moving everything.

**The standard-library patch is gone.** Four characterization tests used to
replace `asyncio.create_subprocess_exec` **process-wide** to observe the spawn.
They now patch the module-level runner seam. That costs the claim that the
parent reads the child's stdout, so the claim is re-owned by two `subsystem`
cases driving a **real** child process, where no fake can satisfy it. Recorded
as a hand-over rather than absorbed silently.

### Defects fixed

| Defect | Evidence |
|---|---|
| A pool slot acquired and then abandoned by an early raise was never returned | Reproduced before and after: queue `1 -> 0` stranded, now `1 -> 1` |
| The restart path could queue **one slot twice** when its replacement acquire failed — `ChromiumPool.release` is an unconditional `queue.put` with no membership check | New test; falsified by restoring the old sequencing |
| The pool-restart retry call site was asserted by nothing; a `slot=None` mutation there survived the entire existing battery | New test; falsified two ways |
| `use_pool = False` was a dead write | Removed |

The pool's own half of the double-release problem — `release` accepting a slot
it already holds — is handed to the step that owns `ChromiumPool`, in that
step's entry.

### One deliberate signature change

The runner resolves its timeout budget from the `env` argument it is handed
rather than from `os.environ`, so its inputs equal its signature. Observationally
identical at the production call site, and it stops the parent's ambient value
silently overriding the environment a later subsystem test builds for a fixture
child.

## Verification

- **Suite 467 -> 485 passed, 2 skipped, 16 subtests, 0 failed** on Linux. Every
  pre-existing node id still collected and passing; the baseline ledger guard,
  which re-runs the suite in a child and diffs node ids, is green.
- **24 mutations, 24 caught**, across five batteries, each run under **both** a
  cleared environment and one exporting every variable the code reads. A further
  **two** were discarded rather than counted: both were no-ops that changed no
  execution and so "survived" while proving nothing. That is why the batteries
  print what they mutated rather than a score — a survivor is the only mutation
  result needing interpretation, and the cheap reading ("the test is weak") is
  wrong often enough to send an author writing tests for a gap that is not there.
- `mypy` clean on its configured target.
- The review system's own 482 tests pass, including the guard that keeps
  internal identifiers out of this public repository.
- Not executed on Windows. The two real-child cases are written to be portable
  (payload written through `sys.stdout.buffer`, child inherits the parent
  environment), and the per-platform green claim belongs to the milestone step
  that takes a real Windows run. Stated rather than assumed: the one previous
  Windows failure here was a test written after the measurement and never run
  there.

## Documents updated

`.system_design/TEST_SUITE.md` (§3.1, §10.4, §11.2), the implementation plan
(this step plus the seven steps that inherit from it), `.coveragerc-gate`, and
both reviewer rule files — the guard registry gains the boundary guard, and the
browser-area rule gains the loader/runner split and why it is not stylistic.

## Review

Reviewed before opening, by both the design reviewer and the code reviewer.

The code review returned **APPROVE** after independently verifying the moved
block is byte-identical, the closure conversion has exactly three intended
substitutions, and the release contract holds on seven paths probed against a
**real** `ChromiumPool` — normal success, non-retryable failure, retry success,
cancellation, a raising `terminate()`, a re-acquire returning `None`, and a
re-acquire raising. It ran its own eleven mutations.

Its one substantive suggestion is taken and is a real fix: `slot = None` sat
*before* `pool.release(stale)`, so a release that raised stranded the slot — the
defect the surrounding change exists to remove, reintroduced one statement
later. Moved after the release; all three orderings of that block are now pinned.

The design review's blockers were documentation: the plan's "Deferred" block
still said this work was undone, two handovers were prose rather than verify
clauses, and the claim that the import guard "is the claim rather than a proxy"
was overstated — `os.posix_spawn` and friends escape it. All corrected. The
second commit is those answers.


shelpuk:
## Answering the automatic review — all four addressed in `a91876e`

Four suggestions, no blockers.

### Three confirm hand-overs that are already recorded — no change needed

- **`WorkerProcess` annotation absent** — deliberate, and the reviewer found the note where it was left. It is E2-4's, and this PR raised that step's urgency: with the four loader tests off the process double, `FakeWorkerProcess` has no production consumer, so the annotation is what restores a *checked* link.
- **Thirteen of fourteen moved helpers carry no docstring** — deliberate, for the reason the reviewer quotes back: an extraction that also rewrites what it moves cannot be reviewed as an extraction. They had no docstring where they came from either. E2-4 owns closing it, alongside annotating the same signatures.
- **~2.0 s heartbeat padding on diagnostics-enabled runs** — pre-existing, moved verbatim, measured at exactly 2.00 s. One correction: it is recorded against **E7-2**, not E7-3. E7-2 owns the lifecycle and will pay it on every hanging-child case.

That the reviewer located all three notes is the point of writing them; nothing to do.

### The fourth is new, correct, and I verified it rather than taking it on faith

**`_terminate_process_tree` does not kill a tree on POSIX.** Probed against a child that had itself spawned a grandchild:

```
before: child alive=True   grandchild alive=True
after : child alive=False  grandchild alive=True  PPid: 1  -> ORPHANED
```

`Process.kill()` signals the direct child only. The Windows branch walks the tree with `taskkill /T /F /PID`, so one function name means two different things by platform. In production the grandchild is **Chromium**, and a SIGKILLed worker runs none of its own cleanup — so a timed-out worker leaves a browser and its profile directory behind on Linux and does not on Windows.

**Not fixed in this PR, deliberately.** The function moved verbatim; changing process termination is a behaviour change an extraction must not smuggle in, and it needs the lifecycle battery to prove it rather than a unit assertion.

Recorded against **E7-2**, whose verify clause already says *"killed parent leaves no orphan"* — that bullet starts red, and the plan entry now says so, with the measurement and two candidate fixes to measure (`start_new_session=True` plus `os.killpg`, or enumerating children before the kill).

The reviewer's framing is the sharpest part of it: the code did not change, but the *name* became more prominent by moving into a module named after process management, and "tree" is exactly the word a future author would trust without checking. That is why it is written down rather than left as a known-known.

Suite still 485 passed, 2 skipped, 16 subtests. Review system's own 482 tests pass.

shelpuk:
Fixed in `59d5416`. Fair catch, and worth stating plainly: the count was wrong on arrival, and it was wrong *inside the sentence explaining that counts go stale*. I added the seventh case answering an earlier review and did not re-count.

Verified the rule the sentence actually exists to state — every case overriding the reuse pin must also double `get_chromium_pool`, or it reaches a real browser: **7 of 7 do.**

So the number is deleted rather than corrected, which is what that sentence was arguing for in the first place. The docstring now records that an earlier wording gave a count and got it wrong, so the next author has the reason and not just the rule.

shelpuk:
Corrected in `0dd3c5e`. Good catch, and the right one to make loudly.

**Neither figure was right.** Re-derived from the batteries themselves: **24** mutations run and caught across five batteries, plus **two** discarded as no-ops — mutations that changed no execution and so "survived" while proving nothing. The plan said 20 (a stale figure from before the review rounds added five more) and the PR body said 25 (an arithmetic slip). Both now say 24, and the PR description is updated too.

Worth naming what this is: it is exactly the failure this repository already has a rule about. The baseline ledger carries separate `Result:` and `Remaining:` fields *because* a number in prose that nothing checks drifts — and I committed the same error twice in one PR, once in a sentence arguing that counts go stale and once in a step whose own notes argue for reporting *what was mutated* rather than a score.

So the entry no longer leads with a total. It names the two discarded mutations, and states the rule they taught: before recording a survivor, name the execution that differs — if you cannot, the mutation is broken, not the test.

PR #69, opened 2026-09-03
Add the SearXNG-contract fixture server and its calibration (plan step E3-2)

shelpuk:

This server's whole job is web search, so our most valuable end-to-end test is
the obvious one: build the wheel, install it, start the real server as a separate
process, and drive it the way a real client would. That test needs a search
engine to answer it — and we cannot mock one across a process boundary. The test
process cannot reach into the server process to patch anything, and we have
deliberately refused to build a "test mode" back door into shipped code, because
this server is unauthenticated and a back door is a bigger risk than any test is
worth.

The way out is that one of our six search providers, **SearXNG, is configured
purely by a URL**. So this PR builds a small local program that answers like a
real SearXNG instance, and the wheel test will point the server at it with one
environment variable. **Configuration, not patching.** No production code
changes; this is test infrastructure only.

Concretely, this unblocks the deterministic MCP-session test — the one that
proves the packaged product actually starts, speaks the protocol, routes a search
and returns an answer. That test could not be written before this.

Two things worth surfacing because they are the kind of mistake that ships
quietly:

- The stand-in was built by reading **SearXNG's own source**, not from memory or
  from tutorials. That caught two errors an assumed contract would have baked in.
  A stand-in that is wrong is worse than none — it makes a broken product look
  tested.
- The stand-in is deliberately **stricter** than the real thing in one place and
  smaller in several. Both are written down, with the reasoning, so nobody later
  mistakes a trade-off for an oversight.

---

## What this adds

`tests/fixture_servers/searxng_contract.py` — a standard-library-only HTTP server
that implements the SearXNG JSON search contract on an ephemeral loopback port,
serving a result set the caller supplies, including an empty one.

`tests/test_searxng_contract_server.py` — its calibration. Seventeen cases:
fourteen drive the server directly over a real socket, three drive **production**
(`search_searxng` and `search_web`) against it, which is what the step's verify
clause asks for.

Documentation: a new §5.2b in `TEST_SUITE.md` carrying the interface, its
rationale and its known limits, next to the sibling fixture-child section; §5.4
and §6.1 amended where this step falsified or extended them; the plan entry
records what landed; the automatic reviewer's rule file gains the guard row.

**Plan step E3-2.** No production code is touched.

## The contract was transcribed, not assumed — and it mattered twice

Read from `searx/webapp.py`, `searx/webutils.py` and
`searx/result_types/_base.py` on `master`.

1. The response envelope has **no `number_of_results`** key. Older documentation
   and several third-party write-ups still show one. A fixture built from memory
   would have served a field the real thing stopped producing.
2. A result carries the **fifteen** keys `LegacyResult.__init__` sets, not the
   three our parser happens to read. An earlier draft served eleven. The design
   document's specimen guard is what caught it — which is the guard doing exactly
   the job it exists for.
3. The `format` parameter is resolved **before** the query is looked at, and an
   *unknown* format is coerced to `html` rather than rejected. So "no `format`
   parameter" yields a web page, and only a known-but-disabled format yields
   `403`. The first draft's blanket "non-JSON format is 403" was wrong.

## Three decisions a reviewer should push on

**The readiness handshake is satisfied by construction, not by polling.**
`socketserver.TCPServer.__init__` binds *and* listens, so the wait §5.4 asks for
has already succeeded when the constructor returns. Measured: a connection opened
before `serve_forever()` was ever called was answered `200`. A poll loop here
would be waiting out a window that does not exist, and its timeout would become
the only number anyone ever measured. §5.4 now records when that substitution is
legitimate — only where the artefact's own constructor listens — so it is not
read as licence to skip the rule for a spawned child, which has a real window.

**The production cases clear the environment outright** rather than deleting the
six provider variables the step's clause names. Production builds its
`httpx.AsyncClient` with `trust_env` at its default, so an ambient `ALL_PROXY`
routes a request aimed at loopback through a proxy. Measured: with the sweep
weakened and `ALL_PROXY=http://127.0.0.1:9` exported, **all three** production
cases fail. With `SERPER_API_KEY` exported instead, the two router cases fail and
the provider case correctly does not. This repository already carries two
separate defences against that same trap elsewhere.

**The `403` branch is deliberately not re-proven here.** It is already pinned one
layer down against a `MockTransport`, which is the lowest layer that can prove
it. Driving it again over a socket would need a fixture knob whose only user is a
duplicate test, and it would land on a weaker assertion anyway: `search_searxng`
catches every per-instance exception and re-raises an aggregate, so the branch's
own message survives only as a substring. The fixture still *serves* the `403` —
the contract table drives it — because that is a fidelity claim about the
fixture, not a claim about production.

## Falsification

Five mutations, each chosen to disarm a claim that would otherwise look proven.
Every one killed exactly the intended cases and no others:

| Mutation | Result |
|---|---|
| Fixed port instead of `0` | the two-instance case fails, alone |
| Teardown moved off `finally` | the failure-path case fails, alone |
| Environment sweep weakened, `SERPER_API_KEY` exported | the two router cases fail; the provider case correctly does not |
| Environment sweep weakened, `ALL_PROXY` exported | all three production cases fail |
| The instance records nothing | the four cases resting on the record fail; the ten that do not, pass |

Control for the last two: sweep intact, both hostile variables exported, all three
pass.

## Checks run locally

CI here runs only the review system's own tests, so these were run by hand:

- Full suite **525 passed, 2 skipped, 16 subtests** — baseline was 508/2/16, and
  the failing set is unchanged (empty) before and after.
- `mypy` clean; `ruff check` clean on both new files (the sibling step's files are
  clean too, so that is the bar).
- The review system's own 482 tests: OK. `scripts/check_plan_dag.py`: exit 0.
- Teardown measured against a handler deliberately sleeping 30 s: `shutdown()`
  0.5001 s, `server_close()` 0.0001 s. That 0.5 s is the stdlib default
  `poll_interval` and is the whole cost of every teardown, so the fixture passes
  `poll_interval=0.05` — the calibration module went 7.35 s to 0.93 s.

## Reviews already applied

The design reviewer raised three blockers before any of this was final: the
result-key count, the unswept proxy and `SEARXNG_*` environment, and the
readiness-handshake rule being negated rather than reconciled. All three are
fixed above. A fourth finding — that `ThreadingMixIn.server_close`'s untimed
`join()` could hang teardown — was **checked and is wrong**: `_Threads.append`
returns early for daemon threads and `ThreadingHTTPServer` sets
`daemon_threads = True`, so that list is always empty. Measured, and recorded in
§5.2b, because the obvious reading of that code says otherwise.


shelpuk:
## Code review answered — four findings, all measured before acting

One of them found a real hole rather than confirming a stated claim. Pushed in `2033a0d`.

### The per-test timeout did not reach the three cases that drive production

Confirmed independently before fixing. `search_searxng` reads `SEARXNG_TIMEOUT_SECONDS` and passes the result to `client.get(timeout=…)`. With that variable cleared — which the hermetic sweep requires — the value is an explicit `None`, and an explicit `None` **overrides** `AsyncClient(timeout=30)` rather than deferring to it. Measured on httpx 0.28.1 against a handler sleeping 20 s:

```
client default (3s):     raised ReadTimeout after 3.09s
explicit timeout=None:   STILL RUNNING at 5s -- no bound at all
```

Those three cases now wrap the await in `asyncio.timeout(HTTP_TIMEOUT_SECONDS)`. Falsified: with the fixture handler wedged for 30 s they now fail in **31 s total** instead of hanging. The constant's comment, which claimed one ceiling covered every call, is corrected — and §5.2b now records that §5.4's bound needs two mechanisms here.

### Pinning each row without pinning the set let a deletion pass — a real hole

The doc-driven guard asserted only that it had parsed *something*. Reproduced: removing the `/elsewhere` row from the design document left this module reporting **16 passed**. That is the exact cheap escape the reviewer rule file names for this guard, and nothing mechanical held it. The row count is now pinned, and the house rule — "pin the SET as well as each member" — is recorded as applying to a driven table too.

### Two claims were false and are corrected, not softened

**"An implementation nobody documented turns CI red"** — false. Verified: adding an undocumented `/healthz` returning 200 left the module green. The guard compares the fixture against the table row by row and cannot see a behaviour the table never mentions. Both the docstring and §5.2b now say so plainly, with the measurement, because a reader who believed the old wording would skip documenting the next route they added.

**"The record-before-respond ordering is asserted indirectly"** — false. Measured myself: moving the record after the response write survives **5 runs of 5**. The window between a client's last read and the handler thread's next statement is too narrow for a synchronous case to land in, and closing it deterministically would need a fixture knob whose only user is that test. So it is now triaged in the docstring of the case that would have to kill it, with the number, per this repo's rule for an unkillable mutant. The lock around the append is triaged the same way.

### Suggestions taken

- **`score` was invented and inverted an upstream invariant.** It served `float(position)` — ascending — where upstream is `weight / position` and `get_ordered_results` sorts descending, so a real instance never returns ascending scores. In a module whose whole claim is transcription, that mattered. Now `1.0 / position`; the specimen's single result stays `1.0`.
- **The documented `{"error": "No query"}` body was prose nothing checked.** The table-driven guard reaches status and content type only. It has its own case now — 18 cases, was 17.
- **The import guard swept one module, not the package.** The consuming job imports `tests.fixture_servers.…`, which runs `__init__.py` too. It now globs the package.
- **`Thread.start()` moved inside a `try`.** The socket is already bound and listening when the constructor returns, so a raise there would strand it — the same shape the sibling step's review found in a process spawn.
- **The specimen query is percent-encoded**, and `_design_section` now names the heading it could not find instead of raising out of `list.index`.
- **§5.2b's path advice was incomplete and would have misled.** It told the packaging job not to copy the `sys.path.insert(… / "src")` line, but `tests/conftest.py` inserts it for anything collected under `tests/` regardless. That is a known open defect owned by the wheel-harness step, and §5.2b now says so rather than implying the path is clean.
- Project memory corrected for the guard count and the `_section_body` coupling count.

### One finding disputed, with evidence

Not from this review — from the design review that preceded it, and repeated here because the code reads as though it were true. The claim was that `ThreadingMixIn.server_close`'s untimed `self._threads.join()` can hang teardown on a wedged handler.

It cannot. `_Threads.append` returns early for a daemon thread and `ThreadingHTTPServer` sets `daemon_threads = True`, so that list is always empty; and `serve_forever` runs the accept loop only, so a stuck handler does not delay `shutdown()` either. Measured against a handler deliberately sleeping 30 s: `shutdown()` **0.5001 s**, `server_close()` **0.0001 s**, serving thread dead, listening socket `fileno() == -1`. Recorded in §5.2b precisely because the obvious reading says otherwise. That 0.5 s is the default `poll_interval` and is the whole cost of every teardown, which is why the fixture passes `poll_interval=0.05` — 7.35 s to 0.93 s across the module.

### One measurement I withdrew

I reported a flake earlier in my own testing: one run in twenty failed because the server returned a `number_of_results` field. It was not a flake. The reviewer was mutating the fixture in the same working directory at that moment to check whether the specimen guard could fail — the timestamps and the choice of key both fit, and it has not reproduced in forty-seven runs since. The guard was working exactly as intended. The methodological error was mine: a flakiness measurement taken while another process is editing the tree is not a measurement. Re-run cleanly afterwards: **20 runs, 0 failures**.

### Gates

Full suite **526 passed, 2 skipped, 16 subtests**. `mypy` clean. `ruff check` clean on all three new files. Review system's own 482 tests OK. `check_plan_dag.py` exit 0. Eight mutations run in total across this branch; seven killed exactly the intended cases, and the eighth is the triaged survivor above.

shelpuk:
## Round 2 answered — all three findings were real, fixed in `6ac43f7`

### Warning: §6.1 still prescribed the delete-list this step measured to be insufficient

Correct, and this is the one that mattered. §6.1 told its author to start the child with `SERPER_API_KEY`, `SERPBASE_API_KEY` and `TAVILY_API_KEY` cleared — the exact prescription §5.2b measured to be inadequate two sections earlier. Anyone implementing the product test from that paragraph would have written a test that fails on a developer's machine for a reason the paragraph made invisible.

It misses, concretely: two provider variables that **already exist** (`SOFYA_API_KEY`, `YDC_API_KEY` — the list belongs to `PROVIDERS`, not to prose), every provider added after this, the eight `SEARXNG_*` tuning variables `search_searxng` reads at the point of use, and the proxy variables that decide whether a request aimed at loopback reaches loopback at all.

§6.1 now says to build the child's environment **additively** — start from nothing, add what the run needs — and points at §5.2b for the measurement. A step that falsifies a sentence owes the edit to that sentence, not only to its own section.

### Suggestion: the `package` job does not exist yet

Correct at all three sites (§5.2b, the fixture module's docstring, the requirements spec). "It **is** imported by a test in the `package` job" is a present-tense claim about something unbuilt — E3-5 and E8-1 build it. Reworded to say the rule is enforced *ahead of* its consumer, with the reason, so the guard does not look like it is guarding nothing: a dependency added in the meantime would otherwise be found by whoever builds that job rather than by whoever added it. "How the `package` job **will** reach it" likewise.

### Suggestion: the measured counts went stale

Correct — the no-query case took the module from 17 to 18 and I did not revisit the figures recorded against it.

**Every figure was re-taken against the module as it stands, not adjusted by arithmetic.** This repo's §1.1 rule — baselines are quoted from real runs or not quoted — applies to a mutation's reported count as much as to a suite baseline, and incrementing a recorded measurement to keep a document self-consistent is the specific failure that rule exists to prevent. Re-run just now:

| Mutation | Re-measured |
|---|---|
| Documented table row deleted, count pin removed | **17 passed** (was recorded as 16) |
| Same deletion, count pin in place | fails at **collection** — `parsed 4 request rows`, expected 5 |
| Undocumented `/healthz` returning 200 | **18 passed** — the module's full count |
| Record moved after the response write | **18 passed**, 5 runs of 5 |

Two of those are now stronger statements than what I first wrote. The pinned-count failure happens at *collection*, not at assertion — louder than described. And the two survivors sit at the module's **full** count, which is the sharper way to say nothing anywhere notices.

I also promoted the record-ordering survivor into §5.2b itself. It was triaged in the test docstring but the design section still implied the ordering was protected; it now states plainly that no test holds it, with the measurement and the reason a deterministic kill would cost more than it is worth.

### Gates

Full suite **526 passed, 2 skipped, 16 subtests**. `mypy` clean. `ruff check` clean. Review system's own 482 tests OK. `check_plan_dag.py` exit 0.

shelpuk:
## Round 3 answered — the standing finding was real, and it was worse than flagged

Fixed in `7211cf0`.

### I fixed three sites, but not the three that were named

The round-2 finding named three sites. My response said "correct at all three sites (§5.2b, the fixture module's docstring, the requirements spec)" — and `git blame` confirms the reviewer is right: `tests/test_searxng_contract_server.py:452` is untouched since `90469b7`. The requirements spec was a fourth site I found myself; I enumerated what I had fixed and presented it as the list I was given, without ever diffing the two. That is a reporting error, not just a missed edit, and it is the part worth correcting: a count is not a check.

### The site I missed was carrying a second uncorrected claim, and a worse one

The docstring of the guard that enforces the stdlib-only rule still read:

> An `httpx` import — or anything under `src/` — would drag a checkout-resolved import into the one job whose purpose is proving the wheel stands alone.

That is the argument **round 1 established to be false**: `httpx` is a runtime dependency of the wheel and resolves from `site-packages` in that job, so it says nothing about the checkout. Only an import of *this project* does. I corrected it in the fixture module and in the requirements spec in round 1 and left it standing here.

Of the worst places for it to survive, this is the worst: a reader asking why the guard exists goes to the guard's docstring, and found the reason that does not hold. Every other mention in the tree — §5.2b, the fixture module, the spec — explicitly *disclaims* that argument. This one still asserted it.

Both errors are fixed in one rewrite: circularity first (an instrument that imported the parser would pin the contract against the code under test), the disclaimed argument named as disclaimed and explained, the tense corrected, and the reason the rule is enforced ahead of its consumer.

### The sweep is in the commit this time

Round 2's warning was the same pattern — correcting a claim where I noticed it rather than everywhere it lives. So rather than fix the named line and stop:

```
$ grep -rn '`httpx` import' tests/ .system_design/ .requirements/
tests/fixture_servers/searxng_contract.py:25:   The argument this deliberately does **not** rest on ...
.system_design/TEST_SUITE.md:615:               ... is that an `httpx` import would break the `package` job. It would not
tests/test_searxng_contract_server.py:462:      The argument this does **not** rest on ... It would not.
.requirements/…/REQUIREMENTS.md:82:             The argument this deliberately does **not** rest on ...

$ grep -rn 'package` job installs\|imported by a test in that job' tests/ .system_design/ .requirements/
none remain
```

Four surviving mentions, all four disclaiming the argument rather than asserting it. Zero present-tense claims about the unbuilt job.

### Gates

Full suite **526 passed, 2 skipped, 16 subtests**. `mypy` clean, `ruff` clean, review system's 482 tests OK, `check_plan_dag.py` exit 0.

No code changed in this commit — one docstring.

PR #82, opened 2026-09-05
Reap a process tree the harness was never told about (plan step E3-4)

shelpuk:

This server fetches web pages by launching a real Chromium browser in a child process. Tests that exercise that path have to start real processes and real sockets, and those are the tests that go bad quietly: they sleep instead of waiting for a signal, they collide on ports, and when one fails it leaves a browser running on the machine. Enough of those and the process table fills up, and unrelated jobs start failing for reasons nobody can trace back.

Nothing in this repository offered a shared way to do that correctly. One test module had a deliberately minimal version of part of it, written for itself, and a second module was already borrowing two pieces across a file boundary with a comment saying "this belongs in the shared harness when someone builds it".

**This is that harness.** It is test-only — no shipped behaviour changes, and no user-facing surface moves. Three later pieces of work are blocked on it: testing the server over a real socket, testing the browser worker's lifecycle, and testing the browser pool.

**The one design decision this step existed to settle.** In production the process tree is worker → browser → renderers, and only the worker announces itself. Our fake worker, until now, spawned exactly one child and *told* the test its process ID. A cleanup routine built on that announcement is never actually made to walk a tree — so "the process tree is gone" could be reported as satisfied by code that could not have found the browser. The harness is now handed one process ID and finds the rest itself, and the fake worker can build a chain of descendants it does not announce, so the walk is exercised rather than assumed.

**The safety property worth knowing about.** A test process shares its process group with everything it starts, including the test runner. The obvious way to kill a process tree — signal the group — would kill the test session, the terminal, and the report that would have told you what went wrong. The harness sends no group signal anywhere. Eleven of its cases run with **no process in existence at all**, driving the kill logic against injected values, precisely because the mistake you most want to catch here is one that produces no output when it happens.

**Cost and confidence.** The suite goes from 785 to 825 tests, and from roughly 130 to 250 seconds. Every rule was checked by deliberately breaking it: 36 distinct mutations, no survivors — though **eleven** of those survived their *first* form, and those eleven are where the real findings were. Both platforms were run.

---

## What is in it

- **`tests/harness/anti_flake.py`** — readiness handshakes (a marker line on a child's stderr, or an accepting port), ephemeral port reservation, portable process-liveness probes, a parentage-walking tree reaper, and a context manager that spawns a child, drains both its pipes, bounds it, reaps its whole tree, and attaches the child's own output to any failure raised inside the block.
- **`tests/child_processes/worker_child.py`** — three new flags: `--grandchild-depth N` (a chain whose deeper generations announce nothing), `--chain-timeout` (so the failure path is drivable), and `--stdout-bytes N` (a payload larger than a Windows command line can carry).
- **`tests/test_anti_flake_harness.py`** — 31 cases calibrating the harness.
- Two existing modules take the two platform probes from the harness instead of defining or borrowing them.
- `TEST_SUITE.md` §5.4a and §5.4b, §5.2a, the implementation plan's landed notes, and the reviewer's guard registry.

## Verification

| | |
|---|---|
| Linux | 825 passed, 2 skipped, 16 subtests, **0 failed** |
| Windows (`windows-latest`, throwaway probe branch, deleted) | 824 passed, 3 skipped, 16 subtests, **0 failed** — the one difference is a POSIX-only case that skips there |
| Collected test IDs vs `main` | 40 added, **0 removed** |
| Mutation testing | 37 mutants, no survivors — 11 of them only after the case or the mutation was reshaped |
| Automatic PR review | **14 rounds, 29 findings, all addressed** — including 11 withheld as out of scope, one of which was a live defect in the harness. Round 14 reported no findings and withheld none, having read the executable code end to end |
| Stability | both new modules, 10 consecutive runs under 12 spinner processes on 4 cores: 0 failures |
| `ruff` on changed files | clean |

## Four things reviews caught that a green suite could not

**A length check standing where a byte-for-byte comparison was claimed.** A pipe pump returning the right *number* of wrong bytes passed every case in both modules. Fixed, and the fix immediately exposed a race the vacuous assertion had been hiding — a failure report was being read from the whole exception note, which also quotes the command line, so `"alpha" in note` was satisfied by the flag that *asked* for the frame whether or not anything was captured. Reading it from the captured-stderr section alone made the case real, and then it flaked 4 runs in 10 under load, because the frame need not have been queued yet. Both are fixed; 0 failures in 10 loaded runs now.

**A pid-reuse window that a docstring said was closed.** The deadline watchdog guarded itself with the check CPython puts inside `Popen.send_signal`, and this branch claimed that made it safe. It does not: CPython checks and signals with nothing in between, while this walks `/proc` between them — about 8 ms — so a caller that reaps in that window frees the process ID before the signal lands. Measured on the unmodified code: **13 of 120 runs** signalled an already-reaped ID. Every reaping call now goes through one lock.

**A fix that shipped with no test, which is the sharpest one.** The lock in the paragraph above was added, documented in three places, and held by nothing: dropping it from the poll, dropping it from the watchdog, or restoring the obvious blocking wait each left all 49 cases in the two modules passing. A measurement showing a fix was *needed* is not a case holding that it is still *there* — and the measurement felt enough like evidence to hide the gap. Two hermetic cases now hold it, plus a third with a real child that specifically pins that the lock spans the walk and the signal rather than the guard alone — a guard-only lock passes the other two.

**A Windows reordering that only a Windows run could see.** Responding to a review comment, I reordered the Windows reap to kill the root and then sweep the tree. It reads more naturally and it leaks the entire chain, because `taskkill /T` finds descendants through the parent ID in the process snapshot and can only walk while the root is alive. The re-run failed two cases with "generation pid N outlived the tree kill". No Linux case can observe it. This is why the platform re-run is a rule rather than a courtesy, and §5.4b now says so.

## Deliberately not done

- **The fake worker's own tests do not run through this harness**, though the step's description reads as though they should. That module is what makes the script a fixed point, and the harness's own tests drive that script — routing it back would leave two instruments measuring each other, and a failure unable to say which one broke. It takes two platform probes from the harness and nothing else.
- **No helper for isolated per-test directories.** pytest's `tmp_path_factory.mktemp` already is one; a fourth would be an unused helper, which this repository has a standing rule against.
- **No process-group kill**, at the stated cost that a descendant started *after* the walk is not reached. Production closes that with a group pass; a test cannot, for the reason above.


shelpuk:
## Both automatic review rounds addressed

**Round one** (against `a4b0569`) raised three; all were already fixed in `8ab4a38`, which landed while it was reading — the duplicated `TEST_SUITE.md` §5.4b block, the §5.4a "nine and twenty-two" breakdown, and the redundant disjunct in the watchdog-disarm case.

**Round two** (against `bc8cc6e`) raised three more. All three were real and all three are fixed in `c3ee481` — but one of them was overstated, and the correction is the interesting part.

### Fixed as reported

- **Module docstring said "Four cases are hermetic".** Thirteen node ids are, across ten functions. Corrected, and the same count was already re-measured rather than incremented in §5.4a and in the guard registry.
- **`test_the_deadline_watchdog_does_not_outlive_the_block_it_bounds` asserted membership where its comment claimed exactly one signal.** Now `assert signalled == [child.proc.pid]`. The comment was right and the assertion was the weaker of the two: membership also passes against a watchdog that had fired early and added a second pid, which is precisely what the case exists to rule out.

### Fixed, but not for the stated reason

- **"Killed chain-armed fixture children leave their chain-record directory behind, every run."** Measured, that is not what happens. The `shutil.rmtree` runs as soon as the chain is complete — not at process exit — so a child killed *later*, which is what every reaping case does, has already cleaned up. Probed directly: the two reaping cases leaked **zero** directories even with the original `tempfile.mkdtemp`.

  The window that *is* real is a child killed **during** the chain wait, before the removal is reached. Probed at **5 of 10 runs**.

  So the fix stands, with a smaller claim attached to it: the record directory now lives beside the caller's `--pid-file`, which every chain user in this suite passes, so pytest's `tmp_path` removes it whenever and however the process died. A hand invocation with no pid file keeps the window, and the docstring says so with the measured numbers rather than the stronger sentence.

  Recording the distinction because "every run" and "5 of 10 when killed inside a 100 ms window" call for different fixes, and the first would have justified restructuring the teardown rather than moving one directory.

### Re-measured after the change

Windows was re-run against `c3ee481` — the fourth probe on this branch, each one after a review-driven change. **823 passed, 3 skipped, 16 subtests, 0 failed**; the three touched modules 90 passed, 1 skipped. Linux **824 passed, 2 skipped, 0 failed**.

That rule earned itself on this branch: an earlier review comment prompted reordering the Windows reap to kill the root and then sweep the tree, which reads better and leaks the entire chain, because `taskkill /T` resolves descendants from the parent id in the process snapshot and can only walk while the root is alive. The re-run failed two cases with "generation pid N outlived the tree kill". No Linux case can observe it. §5.4b carries the evidence.

shelpuk:
## Round three addressed — and the warning is the most interesting finding on this PR

**`blocking_kill never kills` — correct, and worse than it looks.** That case injects a killer that only blocks, so the hanging child survived the block: teardown expired its wait, both pipe pumps expired theirs, and the interpreter ran on to its five-minute backstop. **31.5 s for a case that needs 1 s**, and one ghost process per full-suite run. Fixed by killing after release; measured 31.55 s → 1.05 s, and the guard-only-lock mutant still dies, so the case did not lose its subject.

The reviewer's last sentence is the part worth keeping: *"On Windows the teardown is clean only because `kill_process_tree` runs `taskkill /F /T` regardless of the injected killer, which is why the Windows re-runs did not see it."* That is exactly right, and it inverts the lesson this branch had been accumulating. Four Windows probes ran before this; all were green; the leak was on **Linux**. A platform whose behaviour diverges can conceal a defect as readily as reveal one, and §5.4b now says so next to the case where Windows caught the reordering.

**The `poll_under_lock` claim** was absolute where the code is not: the readiness-failure handler's wait is unlocked, and safe only because the watchdog is armed *after* the handshake. That is a property of the arming order rather than of the wait, so it is now written down at both ends — the method names that call as the one exception, and the call site points back — which is what gives a future edit that moves the arming something to contradict.

**The lock-case arithmetic** is corrected to two hermetic plus a third with a real child, in the plan's landed note and in the PR description. §5.4a already had it right; the other two documents did not, and a reader reconciling three numbers should not have to pick.

Windows re-run against `05595c6` — the fifth on this branch: **823 passed, 3 skipped, 0 failed**; three touched modules 90 passed, 1 skipped. Linux **824 passed, 2 skipped, 0 failed**.

shelpuk:
## Round four addressed — documentation only, no executable line changed

All three findings correct, and the warning names a pattern rather than a typo.

**The lock invariant was corrected in one of its two homes.** `05595c6` removed the absolute claim from `poll_under_lock`'s docstring and left it standing in `TEST_SUITE.md` §5.4a — which is the half a reviewer actually reads. Both exceptions are now stated in both places, each with the reason it is safe: the teardown takes the lock *directly* and holds it across blocking calls, because by then the timer is cancelled and joined and nothing waits on a deadline; the readiness-failure wait is unlocked, safe only because the watchdog is armed **after** the handshake — a property of the arming order, not of the wait. Written that way so an edit that moves the arming earlier has something to contradict, which an absolute invariant does not give it.

**Counts.** The landed note said the three flags arrived "with five cases beside them"; four landed. The script's module docstring said "one case loads it"; three call sites across two modules do.

Windows was **not** re-run for this: `git diff -U0` on the only source file touched shows prose lines and nothing else, so the measurement recorded against `05595c6` still stands. Every earlier round that changed behaviour got its own run — five so far.

Linux **824 passed, 2 skipped, 0 failed**. Windows **823 passed, 3 skipped, 0 failed**.

shelpuk:
## Round five addressed — and I have swept the rest rather than wait for round six

**The finding is correct and its framing is the useful part.** `628c204` fixed "one case loads this script" in the module docstring and left the identical claim in `TEST_SUITE.md` §5.2a — plus the singular "the case that checks it" in the paragraph above it *and* in `stdout_pattern_byte`'s own docstring, which the finding's "also at" list points to. All three now say three call sites across two modules: two re-derive the stdout pattern, one re-derives the descendant program the identical-command-line case needs a look-alike of.

**This is the third time on this PR** that a claim was corrected in one home and left standing in the other — the lock invariant (round four), the importer count (round four's third finding, half-done), and this. That is a pattern, not three slips, so the discipline is now written into §5.2a beside the sentence that kept getting it wrong: when a review names a count, `grep` the claim across `.system_design/` and `tests/` before calling it fixed. Editing the file the reviewer pointed at is not the same as editing the claim.

I ran the suggested sweep — five hits, three of them this claim, two unrelated (`test_dependency_constraints.py`'s "one case per tool per extra" and a plan note about a different step).

### And every other cross-document number, re-measured against the tree

Since three of these have now gone wrong, I checked the rest rather than let them arrive one per round. Measured at head, against what each document asserts:

| Claim | Measured | Asserted in |
|---|---|---|
| harness cases | **35** node ids, 28 functions | §5.4a, plan landed note, guard registry |
| of which hermetic | **13** node ids, 10 functions | §5.4a, module docstring |
| harness cases that are load-bearing / calibrating | 2 / **33** | guard registry |
| hermetic cases driving injected values | **11** (13 − 2 load-bearing) | guard registry |
| fixture-child cases | **18** node ids, 16 driving a real process | §5.2a, guard registry |
| cases the harness step added to it | **4** | plan landed note |
| guard-registry rows | **15** | its own count sentence |

All agree. The mutation total (36) and the platform figures are unchanged, since the last three commits touched no executable line.

Linux **824 passed, 2 skipped, 0 failed**. Windows **823 passed, 3 skipped, 0 failed** at `05595c6`, still the newest commit that changes behaviour.

shelpuk:
## Round six addressed — including the four suggestions it withheld

Round six reported **no findings** and confirmed all 13 from rounds 1–5 fixed and verified in code. It then named four suggestion-level items it had withheld as out of scope from this round on. All four were real, so I have taken them rather than let "out of scope" stand in for "not a problem". Three are prose; one is not.

### The one that is not prose

**`observed: 0` rested on a scheduling window.** Correct, and the reviewer's own hedge ("narrow and never observed") understates why it matters. The count was taken from a **fresh read of the record directory after the wait had already given up** — so it described a later moment than the wait did, and could come back equal to `expected` on the path that exists precisely because it was not. A frame reading `{expected: 1, observed: 1}` while announcing the chain incomplete reads as a defect in the script.

Measured both ways:

| | result |
|---|---|
| At normal speed, 80 runs (40 of them under 12 spinners on 4 cores) | `observed: 0` every time — the window is microseconds wide |
| Window forced open with a 1 s delay, **re-reading** form | `{expected: 1, observed: 1}` — self-contradicting |
| Window forced open with a 1 s delay, **fixed** form | `{expected: 1, observed: 0}` |

`_await_chain` now returns what it last saw alongside whether it completed, and `_report_incomplete_chain` is *given* the count instead of fetching one. The reaping still reads fresh — a generation that recorded itself after the wait gave up is still one the script has to kill, and reaping only what the wait happened to see would leak it. What is given up is a hint: a fresh count would have told a reader whose bound was merely too short that the chain finished just afterwards. A count that can contradict its own frame is the worse of the two.

No case can distinguish the two forms at normal speed, so that is recorded as an equivalent mutant in the docstring of the case that would have had to kill it, with the forced-window measurements. What holds it now is construction rather than a test.

### The three prose items

- **"section 5.2a forbids a startup budget"** — cited in three modules, and §5.2a never contained that rule; it lived only in the plan's E3-1 clause, which is not a durable citation for shipped code to point at. The rule is now stated in **§5.4**, where the three modules already believed it was, with its measured rationale, and the citations point there.
- **"forty-nine cases in these two modules"** — three homes, and the tree now holds 53. Kept as the measurement it is and labelled as what those modules held at the time, per this repository's standing rule that a stale measurement is labelled stale rather than silently corrected.
- **`nodriver_worker.py:608`** → the code is at `:611`.

### Re-measured

Windows re-run against `3c2cd7c` — the sixth on this branch, one after each behaviour change: **823 passed, 3 skipped, 0 failed**; the three touched modules 90 passed, 1 skipped. Linux **824 passed, 2 skipped, 0 failed**. mypy clean, `check_plan_dag.py` exit 0, review-system tests green.

I would rather the withheld findings kept arriving than not. If round seven has warnings-or-criticals only by policy, that is understood — but the four above were worth the round trip, and one of them changed code.

shelpuk:
## Round seven addressed — the withheld four again, and two of them caught me repeating myself

No warnings or criticals, and the round confirms `3c2cd7c`'s fix correct and deterministic. Four more withheld suggestions, all real, all taken.

**Two of them are the discipline failing on its own terms**, which is the part worth writing down. Last round I added a rule to §5.2a: when a review names a count, grep the claim across `.system_design/` and `tests/` rather than editing the file the reviewer pointed at. I then ran that sweep for `forbids` and missed two sites saying **`warns`**; and for `one case loads` and missed a docstring saying **`the one case that needs a value from it`**. Grepping one *phrasing* is not grepping the claim. The refinement — match on the **subject** (the section number, the identifier, the figure), not on the sentence around it — is now recorded beside the rule it corrects.

**§5.2a's "the guarantee below" had no referent**, and this one is more than tidiness: two other homes cite §5.2a as where the half-built-tree guarantee lives, including a case docstring calling it "the exact property section 5.2a promises". §5.2a never stated it — only the script's own docstring did. It is stated there now, so both citations resolve. A citation with no referent is how a guarantee quietly stops being one, which is exactly what this branch spent six rounds discovering about counts.

**The mangled sentence** was mine: a mechanical relabel produced "left all every one of the forty-nine cases". Repaired, along with a sibling stale count on the same pattern ("left all eighty-one cases") that the round did not flag.

### No Windows re-run, and that is checked rather than claimed

Every change here is comment or docstring text. Rather than assert it, I compared each changed test module's **syntax tree against the last Windows-measured commit with docstrings stripped** — all three identical, so the measurement at `3c2cd7c` still stands.

Linux **824 passed, 2 skipped, 0 failed**. `ruff` clean on the changed files, `check_plan_dag.py` exit 0.

### On the withheld-suggestions policy

Rounds six and seven each withheld four suggestion-level findings as out of scope, and described them in prose. Eight items, all real; one of them changed code. I would rather they kept arriving. If the policy is fixed, the prose description is enough to act on — this comment and the last are the record that they were.

shelpuk:
## Round eight addressed — the finding is in the paragraph about not making this mistake

The warning is correct and I am not going to dress it up. The paragraph this branch added to §5.2a, explaining that you must grep the *claim* rather than one phrasing of it, had its two examples attributed to each other's sweeps: `forbids`→`warns` was the **startup-budget citation** sweep, not the importer count, and "the case count" named nothing else in the document. The commit message introducing the paragraph had it the right way round; the paragraph did not. A reader applying the recorded discipline to an importer-count claim would have grepped `forbids` — precisely the miss it exists to prevent.

Fixed as suggested. Both exemplars are now **checked against the commits they describe** rather than recalled:

| claim in the paragraph | verified against |
|---|---|
| the startup-budget sweep used `forbids` | `3c2cd7c`'s message names it |
| it missed two sites saying `warns` | `5c3085f` removes exactly two such lines |
| the importer sweep used `one case loads` | `ecf3568`'s message names it |
| it missed one saying `the one case that needs a value from it` | `5c3085f` removes exactly one |

The "eighty-one" sentence, still gapped after the last relabel, now names the three modules that count came from.

### The honest reading of rounds 5 through 8

Four consecutive rounds have found defects in this branch's *prose* rather than its code — a stale count, a citation to a rule its section never contained, a dangling referent, a mangled relabel, and now a self-inverted example in the paragraph about that very failure. The code has been clean since round five. That asymmetry is the signal: this change ships an unusually large amount of explanatory documentation, and dense prose written to justify decisions has a defect rate that the suite cannot see and mutation testing cannot reach. Two things follow, and both are now in the tree rather than in this thread: the discipline paragraph itself, and the practice of checking a documented measurement against the commit that produced it instead of against memory.

### Verification unchanged

Documentation only, and checked rather than asserted — the three changed test modules' syntax trees are identical to the Windows-measured commit `3c2cd7c` with docstrings stripped, so §5.4b's figures stand. Linux **824 passed, 2 skipped, 0 failed**; `ruff` clean; `check_plan_dag.py` exit 0.

shelpuk:
## Round nine addressed — and the discipline paragraph earned itself

**The warning is the third home of a claim rounds three and four corrected in the other two.** The plan's landed note still asserted "Every reaping call now goes through one `poll_under_lock`" with no qualifier, while `poll_under_lock`'s own docstring and §5.4a had both been corrected to name the two calls that deliberately do not — the teardown, which takes the lock directly and holds it across blocking calls, and the readiness-failure wait, which predates the watchdog being armed. Fixed, with a pointer to §5.4a for the reasons.

Worth reporting precisely because it is the failure this branch has been documenting: **applying the discipline found it in one command.** `grep -rn poll_under_lock` across `.system_design/` and `tests/` returns fifteen sites, and exactly one made an unqualified claim. Grepping by *subject* rather than by phrasing works. The paragraph had simply never been applied to a claim that was fixed before it was written — so the note now says it is the third home, which is what tells the next reader to look for a fourth.

**The count.** "Seven things the build changed" is followed by eight bold paragraphs; the sibling notes for E3-1, E3-2 and E3-3 all match their counts to the paragraphs that follow. Counted rather than adjusted by one — the eight are listed in the commit message.

### Where this stands

Nine rounds. The code has been clean since round five: rounds six through nine have found ten items, every one of them in this change's own prose, and the executable diff has not moved since `3c2cd7c` except by that round's own fix. Verified again rather than asserted — the three changed test modules' syntax trees are identical to the Windows-measured commit with docstrings stripped, so §5.4b's figures stand.

Linux **824 passed, 2 skipped, 0 failed**. Windows **823 passed, 3 skipped, 0 failed** at `3c2cd7c`. `ruff` clean, `mypy` clean, `check_plan_dag.py` exit 0, review-system tests green.

I will keep taking rounds for as long as they find things. If the remaining findings continue to be prose residue in documentation this change itself introduced, that is a fair description of where the risk now lives — and it is a better place for it than the harness.

shelpuk:
## Round ten addressed — the withheld suggestion was a real defect, and the best finding on this PR

Round ten reported no warnings or criticals and withheld one suggestion, described in a single line: *"a latent asymmetry in the harness's truncation guard that no current case exercises."* That was enough to find it, and it is a defect rather than prose.

**`stdout_bytes` refused a truncated payload from one kill path and not the other.** The guard exists because readiness is announced *before* a payload is written, so a caller who leaves the block early compares a prefix — it turns that mistake into a sentence. The flag arming it was set by the teardown reap and **not** by the deadline watchdog. So a caller who had *explicitly asked* for the child to be killed mid-flight got back exactly the silently-short payload the guard was written to prevent.

Measured before the fix, with a child that writes in slices under a half-second deadline:

```
deadline_fired      : True
killed_while_running: False
stdout_bytes() RETURNED 3000 of 20000 bytes -- no complaint
```

and after:

```
killed_while_running: True
stdout_bytes() refused: this child was killed while it was still running, so its
stdout may be truncated -- wait for it to exit inside the `with` block ...
```

**Why it survived ten rounds, which is the part worth keeping.** No case could reach it. The fixture child writes its whole payload *before* it hangs, so every kill lands after the last byte — the artefact that every reaping case uses is structurally incapable of being truncated. The new case drives a slice-writing artefact instead, and asserts the **refusal**, not the length: pinning "3,000 bytes" would pin how far a Python interpreter gets in half a second, which measures the runner rather than this code. Mutation-checked — removing the flag from the watchdog kills the case.

### Re-measured, not incremented

The counts moved with the new case, so every home was re-derived rather than adjusted: **36** harness node ids from **29** functions, 13 hermetic from 10 unchanged, 18 fixture cases, 15 registry rows; mutation total now **37**. A sweep by subject found no surviving "thirty-five"/"twenty-eight"/"thirty-three" claim about this module.

Windows re-run against `7caab58` — the seventh on this branch, one after each behaviour change: **824 passed, 3 skipped, 0 failed**; the three touched modules 91 passed, 1 skipped. Linux **825 passed, 2 skipped, 0 failed**.

### On the withheld-suggestion policy, now with a concrete data point

Eleven suggestion-level findings have been withheld across rounds 6–10 and described only in prose. Ten were documentation. **This one was a live defect in the harness** — the kind the policy is presumably designed to let through when a change is otherwise sound. One line of description was enough for me to locate, reproduce and fix it, so the prose summaries are doing real work and I would rather they kept coming. But it is worth saying plainly that "suggestion-level, out of scope" was, this round, a guard that silently returned the wrong answer under exactly the conditions it was written to catch.

shelpuk:
## Round eleven — the finding's arithmetic is wrong, its request is right, and then I repeated the mistake

No warnings or criticals, and the round verified the truncation-guard fix correct, calibrated and mutation-verified. One low-confidence suggestion: that §5.4b's "91 passed, 1 skipped" does not re-derive from the three modules, which the round put at **82** node ids.

**Measured, they hold 92:**

```
tests/test_anti_flake_harness.py         36
tests/test_worker_child_fixture.py       18
tests/test_worker_runner.py              38
total                                    92
```

and the probe's own invocation reports `92 passed` on Linux. The recorded Windows figure — 91 passed, 1 skipped — is exactly that, with the POSIX-only process-group case skipping. So the figure was right.

**The ten-id gap the round noticed is real, and it is the unit.** `test_worker_runner.py` carries 38 collected ids from 33 test functions: ten are `async def`, and two are parametrized into seven ids. Counting functions lands at 33; counting `^def test_` lines lands at 23; counting `def test_` at any indent lands at 33. Every cheaper unit than "collected node ids" lands short.

**The half of the finding that stands is the one I have taken.** The invocation lived only in the probe workflow, which is deleted along with its branch by design — so a reader had no way to re-derive the number without guessing what was run. §5.4b now records the command, the arithmetic, and which unit to count in.

**And then I got it wrong in the same way.** The commit answering this wrote "38 ids from 28 functions" — 28 being `^def test_` alone, missing the ten `async def`. Corrected to 33 in a follow-up, with both wrong numbers and their causes named in §5.4b. I have not rewritten the commit message: history is not the place to pretend a figure was right.

That is twice in one exchange that a count was taken in a unit nobody stated — once by the review, once by me. §5.4b now states the unit at the point of use, which is the only fix that outlives either of us noticing.

Linux **825 passed, 2 skipped, 0 failed**. Windows **824 passed, 3 skipped, 0 failed** at `7caab58`. Documentation only since; the changed test modules' syntax trees are unchanged from that commit with docstrings stripped.

shelpuk:
## Round twelve — correct, and it caught the worst thing on this PR

The finding is right and it is not really about a number.

The paragraph I wrote to correct the runner-module figure said the erroneous **28** was "the count of `^def test_` lines". It was not. That count is **23**, at every commit of this branch and on `main`, and *no* counting method in the tree yields 28:

```
^def test_                    23
^async def test_              10
^\s*def test_                 23
^\s*(async )?def test_        33
def test_  (any position)     33
```

**28 came from arithmetic on the review's own figures — `82 − 36 − 18` — and was then handed a method that would plausibly have produced it, without measuring.** That is worse than the wrong number, and it is the part worth recording. The number is one error. The invented provenance is an error dressed as evidence: a later reader re-derives from the stated method, finds 23, and cannot tell which half to distrust — the figure, the method, or both.

§5.4b now says where 28 actually came from and names that as the defect rather than the count. The commit messages keep their errors; history is not the place to pretend a figure was right.

### The pattern across the last four rounds

Rounds 9–12 have found, in order: a claim corrected in two homes and not a third; an example paragraph with its own examples inverted; a live defect in a guard no case could reach; and now a fabricated justification for a wrong number. Three of those four are failures of *explanation* rather than of code — and this last one is the failure mode that documentation-heavy work is most exposed to, because prose can assert a derivation that was never performed and nothing executes it.

The two disciplines this branch has been accumulating both aim at that, and both are now in §5.2a and §5.4b rather than in this thread: grep the **claim** by subject rather than by phrasing, and state the **unit** at the point of use. This round adds a third, which I will state here and leave for whoever picks up E7-1: **do not reconstruct how you got a number you did not measure.** Say it was unmeasured.

Linux **825 passed, 2 skipped, 0 failed**. Windows **824 passed, 3 skipped, 0 failed** at `7caab58`. Documentation only since — the changed test modules' syntax trees are identical to that commit with docstrings stripped, checked rather than asserted.

shelpuk:
## Round thirteen — correct, and it lands on my correction rather than the count

**28 is a real count.** The runner's non-async collected ids number exactly 28: 23 sync functions, two of them parametrized into seven ids — `23 − 2 + 7 = 28` — with the ten `async def` ids making up the remaining ten of 38. So the previous commit's "no counting method in the tree yields 28 at all" is false.

That makes three wrong statements in sequence, each correcting the last:

1. **28 functions** — wrong; the module has 33.
2. **"28 was the count of `^def test_` lines"** — a fabricated provenance; that count is 23. The number had actually come from `82 − 36 − 18`, and was handed a plausible-looking method without measuring.
3. **"no counting method yields 28"** — an absolute negative drawn from a finite check. Five greps returning nothing does not license "no method exists", and 28 turns out to be a perfectly real count in the wrong unit.

The third is the same defect as the second with the sign flipped: both assert something about a derivation that was never performed. Correcting an over-claim by over-claiming in the other direction is not a correction.

§5.4b now states only what was measured — **38 ids from 33 functions; 28 of them non-async; 23 `^def test_` lines** — and says plainly that **which route produced the original 28 is not known**, because it was not measured at the time. That sentence is shorter than either wrong version and is what should have been written first. Every figure in the paragraph is now checked programmatically against the module rather than recalled:

```
OK  38 collected ids        OK  two parametrized
OK  33 functions            OK  23 - 2 + 7 = 28
OK  23 sync functions       OK  non-async ids == 28
OK  10 async functions      OK  ^def test_ == 23
```

Commit messages keep their errors; history is not the place to pretend a figure was right.

### Where the risk on this change now lives

Thirteen rounds. The executable code has not changed since `7caab58`, and rounds 11–13 all read it end to end without finding a defect in it. Every finding since round nine has been in prose *about* the code — and the last three specifically in prose written to correct earlier prose. That is a real signal about this change: its documentation carries measurements and derivations that nothing executes, so an assertion about how a number was obtained can be wrong in a way no test will ever catch.

Three disciplines have come out of it, all now in the tree rather than in this thread — grep the claim by subject, state the unit at the point of use, and **do not reconstruct how you got a number you did not measure; say it is unmeasured.** The paragraph carrying the third is now itself the worked example of getting it wrong twice.

Linux **825 passed, 2 skipped, 0 failed**. Windows **824 passed, 3 skipped, 0 failed** at `7caab58`, unchanged and re-verified by syntax-tree comparison.

shelpuk:
## Round fourteen: no findings, and nothing withheld — the review has converged

Round 14 is the first round with **no findings and an empty withheld count**. It read the executable code independently rather than on trust — `anti_flake.py` end to end, the `worker_child.py` diff in full, all 29 calibration functions, and the two migrations — and confirmed each load-bearing case would fail if its subject mutant were restored. The last executable change remains `7caab58`; everything since is TEST_SUITE.md prose.

### Final state

| | |
|---|---|
| Linux | **825 passed, 2 skipped, 16 subtests, 0 failed** (from 785) |
| Windows `windows-latest` | **824 passed, 3 skipped, 16 subtests, 0 failed** — one POSIX-only case skips |
| Collected node ids vs `main` | **40 added, 0 removed** |
| Mutation testing | **37 mutants, no survivors** |
| Stability | both new modules, 10 runs under 12 spinners on 4 cores: 0 failures |
| `ruff` / `mypy` / plan graph / review-system tests | clean |
| Windows probe runs | **7**, one after every behaviour change; each branch deleted |

### What fourteen rounds actually bought

Two design-review passes found 11 blockers before a line was written — including two routes by which this harness could have killed a process it never spawned. Two code-review passes then found 15 more, of which the sharpest was that the pid-reuse lock had shipped **documented in three places and held by nothing**: three separate ways of deleting it left all 49 cases passing.

The automated rounds split cleanly in two. Rounds 1–5 and round 10 found **code**: a length check standing where a byte comparison was claimed; a `pytest.fail` that escaped the capture handler; a thread joined before it was started; a temp directory with a real if narrow leak; and a truncation guard that covered one kill path and not the other, handing back 3,000 of 20,000 bytes without complaint. Rounds 6–13 found **prose** — and the last four found prose written to correct earlier prose, ending in a three-deep sequence where a wrong number got a fabricated provenance, and the fix for that was an absolute negative that was itself false.

That asymmetry is the honest summary of where risk lives in this change. The suite proves the code. Nothing executes the documentation, so a claim about *how a number was obtained* can be wrong in a way no test catches — which is why the three disciplines this branch produced are in the tree rather than in this thread:

- **grep the claim by subject**, not by one phrasing (§5.2a);
- **state the unit at the point of use** (§5.4b);
- **do not reconstruct how you got a number you did not measure** — say it is unmeasured (§5.4b), a paragraph that is now its own worked example of getting that wrong twice.

Ready for human review. Not merging.

PR #94, opened 2026-09-11
Add Serply as a seventh search provider

googio:
## What

Adds Serply (https://serply.io) as a search provider, appended last in `PROVIDERS`. Selection for every existing key is unchanged: Serply is picked only when no other provider is configured.

## Why

Serply returns Google results through a plain GET with an `X-Api-Key` header, so it fits the existing provider contract with no router change. It gives operators another keyed option beside Serper and SerpBase.

## Notes

- `search/serply.py` mirrors `serper.py`: one coroutine, reads only its own env var, leaves `raise_for_status()` to the router, returns `page_content=""`.
- A missing or reshaped `results` list raises `SerplyError`, the same choice `sofya.py` and `youcom.py` make, rather than SerpBase's empty list.
- The registry consistency tests drove the doc edits: tool docstring, `.env.example`, the README provider lists, and the review rule's backend count.
- New rows in `test_search_provider_error_paths.py` and `test_provider_credential_disclosure.py`; their docstring counts move from six to seven. `SERPLY_API_KEY` is added to the ledger's cleared names.
- `tests-broad.yml` floor re-measured at 1251 (was 1222) after the new cases, as the comment above it asks.

## Testing

```
uv run pytest -q
1249 passed, 2 skipped, 16 subtests passed in 131.16s
```

Baseline on main before the change: 1220 passed, 2 skipped. I also ran `search_serply` and `search_web` against the live API with a real key. Both returned results, and the router selected Serply with only `SERPLY_API_KEY` set.

Serply support is optional, and behavior is unchanged when `SERPLY_API_KEY` is unset.

Disclosure: I work with Serply. Happy to adjust scope, naming, or drop this entirely if it isn't a direction you want for the project.


shelpuk:
Thanks for this, @googio. I pushed a maintainer commit on top of yours to get the PR ready to merge. Here's what changed and why.

**Request form**
- Serply requests now use the form in Serply's docs: `GET https://api.serply.io/v1/search/` followed by the URL-encoded `q=...&num=...` in the path. This matches the [Google Search reference](https://serply.io/docs/resources/google-search) and [`serply-inc/examples`](https://github.com/serply-inc/examples/blob/master/python/simple_search.py). The query-string form (`/v1/search?q=...`) works, as you tested, but it isn't documented, so we'd rather not depend on it.
- We haven't run the path form against the live API, because we don't have a key. This doesn't block the merge. If you can try a few queries on the updated branch (a plain one, `c++ & rust`, `a/b`, and one with non-ASCII text), that would still help: some servers reject `%2F` in a path.
- `num` is still sent as given. The docstring no longer says how the API handles large values, since Serply doesn't document a limit.

**Tests**
- `tests/test_serply_unit.py` is rewritten in pytest style. New test files in this repo are written pytest-first (TEST_SUITE.md §3.1), so it's also removed from the E11-1 migration batch. Environment changes now go through `monkeypatch`, so nothing leaks between tests.
- New cases:
  - URL encoding for `&`, `+`, `/`, `?`, `#` and non-ASCII text
  - large `num` sent unchanged
  - result order preserved
  - a blank query or `num_results < 1` sends no request
  - an unset or whitespace-only key sends no request
  - the default client uses a 30 s timeout
  - a non-string `description` becomes an empty snippet
- Router tests: the Serply test no longer leaks environment variables, and a new test checks that You.com is still chosen when both `YDC_API_KEY` and `SERPLY_API_KEY` are set.
- A mutation check on `serply.py` (18 small deliberate breakages) was caught by the tests every time.
- A new credential-disclosure test covers very long queries for SerpBase and Serply. httpx rejects the URL with `InvalidURL`, and the test checks that the error message contains neither the key nor the URL.

**Docs**
- `serply.py` now has Google-style docstrings.
- Provider counts that still said six (or five) are fixed in test docstrings, the review rules and TEST_SUITE.md.
- TEST_SUITE.md §3.1 and §14 record the path-form decision and that it wasn't tested live.
- The `tests-broad` floor is now 1267.

googio:
Thanks for finishing this off and merging. I ran the path form from main against the live API as you asked, with the same `urlencode` output the module produces and the httpx default User-Agent. Results:

| Query | Path form (`/v1/search/q=...`) | Query form (`/v1/search?q=...`) |
| --- | --- | --- |
| `rust async runtime` | 200, 3 results | 200, 3 results |
| `c++ & rust` | 200, 3 results | 200, 3 results |
| `a/b` (`%2F`) | 200, 2 results | 200, 2 results |
| `tokio 東京 ランタイム` | 200, 3 results | 200, 3 results |
| 300-char query | 200 | 200 |
| `100%` (`%25`) | **400 Bad Request** from the edge | 200, 3 results |
| `zebra?giraffe`, num=2 | 200 but **10 results about zebras** | 200, 1 result (Okapi) |
| `tokio#zzzz`, num=2 | 200 but **10 results about Tokyo** | 200, 2 results |

So `%2F` is fine, but two things break in the path form:

1. A literal `%` in the query (`%25` after encoding) is rejected with 400 before it reaches the API.
2. `%3F` and `%23` are decoded server-side and then re-parsed as the start of a query string or fragment. The search runs on the truncated text (`zebra`, `tokio`), and `num` is dropped so the default of 10 comes back. The results look valid, which makes this the worse failure mode.

The query-string form returned the full query and honored `num` in every case. It is also the form the other Serply integrations I maintain use, and the API has kept it stable. I would suggest switching `SEARCH_ENDPOINT` back to `https://api.serply.io/v1/search?` with `urlencode` appended, and the new encoding tests you added would carry over unchanged. Happy to open a small follow-up PR with that change and a note in TEST_SUITE.md if you prefer it that way.

PR #99, opened 2026-09-18
fix serpbase

shelpuk:


shelpuk (commented):
## Review: `fix serpbase`

Thanks for this — the change is correct, and I verified the premise rather than taking it on trust. SerpBase's current docs confirm every part of it:

| Your change | Docs say |
|---|---|
| `GET` → `POST` | "POST only" |
| `api_key` param → `X-API-Key` header | "pass your API key in the `X-API-Key` header" |
| `organic_results` → `organic` | `organic` |
| dropped `num` | **no `num` parameter exists** — params are `q`, `hl`, `gl`, `page`, `device` |

That last row is worth calling out because it looks like an omission and isn't: there is no way to ask SerpBase for a result count, so dropping it is right, and `num_results` is still honoured by the existing client-side truncation. `test_returns_at_most_num_results` already covers that, so no gap.

The unit-test rewrite is the strongest part of the PR — asserting method, URL, header and body is a real contract assertion rather than a restated implementation.

**I have opened #102**, which merges this with `main` and fixes what the merge breaks. I could not push here (`maintainerCanModify` is off on this fork). Your commit is preserved with authorship; merging #102 closes this.

---

### The merge is worse than a conflict

The conflict is one prose hunk. The dangerous part is what merges **cleanly and wrong**. This branch predates `serply` and `apifare`, so git combines "serpbase no longer carries its key in the URL" with tests `main` wrote while it did:

```
FAILED tests/test_provider_credential_disclosure.py::
  test_an_over_long_query_fails_without_quoting_the_url_or_the_credential[serpbase]
E   Failed: DID NOT RAISE InvalidURL
```

That test is parametrised over providers that put the query in the **URL**. After your change SerpBase POSTs a JSON body, so its URL is a fixed string that no query can lengthen. Nothing in the diff touches that test, which is why neither side's CI would have predicted it.

The conflict hunk itself is interesting for the same reason — **neither side is right after merging**. Yours says "in the request URL for SearXNG, and in a request header for the other five"; `main` says "the two providers that carry it there". Post-merge there are eight providers and SearXNG is the only one carrying it in the URL. #102 names it rather than counting, which is the convention `main` adopted in b60c884 ("stop counting providers").

### Three places the merge leaves saying something false

None are in your diff — they're collateral, and I mention them so the reasoning is on the record rather than implied by #102:

1. **`.github/review/rules/search-providers.md`** — "SerpBase's `api_key` query parameter and SearXNG's base-URL userinfo" are named as the two credentials reachable through a URL. This is the rule file loaded for *every* `search/**` change, so it would have misdirected the next provider review.
2. **`.system_design/TEST_SUITE.md`** — describes SerpBase and Serply as putting the query in the URL, and §14 describes `serpbase.py` in the present tense as putting `api_key` in `params`. I kept the incident record as written (it is what happened) and corrected only the present-tense half.
3. **`--min-selected`** in `tests-broad.yml` — removing one parametrised case takes the count 1367 → 1366, and `tests/test_ci_collection_floors.py` holds that to **equality**, not "at least". Worth knowing for any future provider PR: it goes red for a reason that never mentions providers.

---

### Findings

**1 — (suggestion) The docstrings trade measured reasons for speculative ones, and one of them didn't need to.**

`SearchProviderTransportError` went from *"at least one provider — SerpBase — authenticates with a query parameter, so the unmodified message carries an API key"* to *"A provider URL may contain credentials now or in the future"*.

The first sentence had to change. But the concrete reason is **still available and still true**: SearXNG authenticates by base-URL userinfo, and its `carried_in_url` flag is `True` in the disclosure table. Replacing a thing that happened with a thing that might happen makes the rule easier for a future maintainer to undo — "may contain credentials in the future" reads like a precaution, and precautions get optimised away. Same pattern in `test_search_provider_error_paths.py`, which also drops the `TEST_SUITE.md` §14 cross-reference to the recorded incident.

#102 restores the concrete form, naming SearXNG and keeping the §14 pointer. Happy to drop that from #102 if you'd rather keep the generic wording — but I'd argue the incident is the reason, and it should stay findable.

**2 — (suggestion) `assert request.content == b'{"q":"apple inc"}'` pins httpx's JSON separators.**

This is the right *claim* — exact bytes prove no extra fields are sent. But it also pins that `httpx` serialises with no space after the colon, and this project declares `httpx[socks]>=0.28,<1`, a range. A patch release changing separators would turn this red for a reason unrelated to SerpBase or to your change.

`json.loads(request.content) == {"q": "apple inc"}` is exactly as strong — it still fails on an extra or renamed field — and is immune to encoder cosmetics. This file already worries about that class of thing: the over-long-query case says it is *"pinned rather than recorded as prose because `httpx` is declared as a range"*. I left this alone in #102 since it currently passes; your call.

**3 — (finding worth its own issue) SerpBase has no live test, and this PR is the moment that matters.**

This rewrites the entire wire contract — method, auth mechanism, request encoding, response envelope — and **not one test in the suite would fail if the documentation were wrong**. The unit tests assert what the code sends against a mock written in the same PR; that is self-consistency, not evidence. My own verification was reading the docs, which is the same class of evidence.

The repo already has the pattern and already describes this exact risk. `tests/test_apifare_live.py`:

> Nothing else in the suite can: `test_apifare_unit.py` proves what is *sent* and what the parser does with a body it is *handed*, but apifare's route slug, its request body and its response envelope are unconfirmed.

Swap "apifare" for "SerpBase" and that is this PR. A `tests/test_serpbase_live.py` mirroring it — `pytest.mark.live`, gated on `KINDLY_RUN_LIVE_TESTS`, structural assertions on the envelope and field types rather than on content — would close it. It needs a funded key to be worth anything, which is why I have not written it into #102; it belongs to whoever holds the key.

Of eight providers only `serper` and `apifare` have live tests, so this is not a bar you are being singled against. But apifare's live test exists precisely because its contract was unconfirmed, and SerpBase's contract just became unconfirmed in the same way.

---

### Verification of #102

- `test_provider_credential_disclosure.py` — 86 passed
- provider + router + serpbase modules — 145 passed
- full broad selection — 1359 passed; the 4 failures are pre-existing on clean `main` on this Windows machine (3 × `test_worker_child_fixture` readiness/pid, plus the ledger case that depends on them)
- `scripts/check_plan_dag.py` exit 0 · `test_ci_collection_floors.py` 35 passed · review-script suite 586

